PR #12, opened 2015-09-15
Filename completion

gwenn:
Ok,
I am not happy with the Competer Trait but it works...
If you have any suggestion, just tell me.
Regards.

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -1,10 +1,13 @@
 extern crate rustyline;
 
+use rustyline::completion::FilenameCompleter;
 use rustyline::error::ReadlineError;
 use rustyline::Editor;
 
 fn main() {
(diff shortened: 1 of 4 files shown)


kkawakam:
Are you unhappy with something specific in the Completer Trait?  It seems sufficient to me

gwenn:
I've tried the following signature (`&mut self`):
```rust
fn complete(&mut self, line: &str, pos: usize) -> Result<(usize, Vec<String>)>;
```
without success.
For example, if you want to make a SQL completer but the schema has changed, you will have to invalidate/mutate the cached database metadata...

gwenn:
The next pull request should be 'incremental history search' (Ctrl-R/S) and may be some 'word' actions (delete previous/next word, ...).
Regards.

PR #13, opened 2015-11-10
Add incremental reverse search

gwenn:


kkawakam:
Looks good to me. Thanks!

gwenn:
The next step may be to handle unknown escape sequences correctly by consuming them all.

Issue #22, opened 2016-02-23
Incorrect prompt width when control chars are used


gwenn:
```
extern crate unicode_width;

fn main () {
    let s = "\x1b[1;32m>>\x1b[0m ";
    println!("{}", unicode_width::UnicodeWidthStr::width(s)); // 12
    let s = ">> ";
    println!("{}", unicode_width::UnicodeWidthStr::width(s)); // should be 3
}
```

gwenn:
I tried `term_grid` crate with no success:
```
extern crate term_grid;

use term_grid::Cell;

fn main () {
    let s = "\x1b[1;32m>>\x1b[0m ";
    let cell: Cell = s.into();
    println!("{}", cell.width); // 12
}
```

kkawakam:
Fixed by PR #23

PR #23, opened 2016-02-24
Fix calculated width of prompt with ansi formatting

gwenn:


gwenn:
Fix #22

examples/example.rs
```diff
@@ -12,7 +12,7 @@ fn main() {
         println!("No previous history.");
     }
     loop {
-        let readline = rl.readline(">> ");
+        let readline = rl.readline("\x1b[1;32m>>\x1b[0m ");
```
kkawakam: Is it really necessary to have this in the example?

gwenn: No, it is not necessary.
It just shows that rustyline properly handles prompt with ansi escape codes.

kkawakam:
Thanks for the fix.

PR #26, opened 2016-03-24
fix for encode_ut8 returning iterator

saghm:
The latest nightly changes `str::char::encode_utf8` takes no arguments and returns an iterator instead of returning a size and taking in a mutable buffer.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -290,8 +290,16 @@ fn edit_insert(s: &mut State, ch: char) -> Result<()> {
     if s.buf.len() < s.buf.capacity() {
         if s.buf.len() == s.pos {
             s.buf.push(ch);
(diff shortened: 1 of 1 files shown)


kkawakam:
Looks good, thank you.  I'll be pushing out a new version of rustyline tonight

kkawakam:
Pushed 0.2.2 to crates.io

saghm:
Actually, when trying to use this now, it seems to be failing with an EOF error every time now. I can't tell for sure if this is from my change, but it seems likely.

I unfortunately don't have any time to delve into it right now, but hopefully I will later tonight/tomorrow. Sorry about the hassle!

kkawakam:
Thanks for taking a look. 

-----Original Message-----

PR #28, opened 2016-03-29
Kill ring

gwenn:


kkawakam:
These changes look good. Just wanted to double check though that M-d & M-DEL are not implemented?

gwenn:
You are right: they are not implemented.

PR #47, opened 2016-07-09
Check winapi calls.

gwenn:
And declare target specific dependencies.


gwenn:
I am cross compiling on Arch Linux so I may introduce some unexpected regression...

kkawakam:
I've cherry picked your commit onto my branch, so i will close this

PR #52, opened 2016-07-15
Parametrize Editor by the Completer type

### LINKED ISSUE #49: Editor taking a trait object reference to Editor is awkward
I saw that windows support is (partially) available for rustyline now so I was considering removing my linenoise-rust dependency and use this instead as the completion support was to primitive for my needs.

Looking at the API however I am going to run into problems with Completer being a borrowed trait object as the current implementation mandates a borrow and prevents Editor from being Sync. This could be solved by instead parameterising Editor by the Completer. 

I'd be happy to implement this and submit a PR if you think it is a good idea.

Marwes:
This allows the `Editor` to implement `Sync` as long as it contains a completer which implement `Sync`. It lets the `Editor` take ownership of the completer.

Fixes #49

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,8 @@ use rustyline::error::ReadlineError;
 use rustyline::Editor;
 
 fn main() {
-    let mut rl = Editor::new();
+    // `()` can be used when no completer is required
+    let mut rl = Editor::<()>::new();
     if let Err(_) = rl.load_history("history.txt") {
         println!("No previous history.");
     }
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -16,7 +16,7 @@ static PROMPT: &'static str = ">> ";
 fn main() {
     let c = FilenameCompleter::new();
     let mut rl = Editor::new().history_ignore_space(true);
-    rl.set_completer(Some(&c));
+    rl.set_completer(Some(c));
     if let Err(_) = rl.load_history("history.txt") {
         println!("No previous history.");
     }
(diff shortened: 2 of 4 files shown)


Marwes:
Happy to update this PR if you find any problems with it. It may be a bit awkward that users need to give a completer type explicitly if they don't use a completer but I am not sure what is the best way to avoid that.

kkawakam:
Hey @Marwes, sorry that I didn't get back to you on your issue during the week.  I agree that with you that it is a bit awkward with explicitly parametrizing Editor if there is no completer but I am okay with it.  Can you update the README.md example to reflect that change?

Marwes:
No problem I am in no rush, just wanted to make sure making the change was possible.

A tip for keeping the README example updated, [skeptic](https://github.com/brson/rust-skeptic) is really useful to ensure that the examples in the README always builds (though it can be a bit slow if you start having 10+ examples as I have found).

Marwes:
:+1:

kkawakam:
@kkawakam note to self: breaking change because of the need to explicitly parametrize the Editor when there is no completer.

Marwes:
@kkawakam I used the pattern of `()` as a 'null object' in this PR which may not be the most intuitive solution. I used this pattern in gluon as well but I have been convinced to change it to an explicit type for the null object instead to increase discoverability (so maybe a `pub struct NullCompleter;` instead).

I obviously don't mind `()` myself but I thought I would let you know that it might not be a common pattern that is easily understandable.

Issue #48, opened 2016-07-09
Moving the cursor wipes out next lines


dcampbell24:
The library calls refresh in the cursor movement functions causing future lines to be wiped out. This is causing issues for a game I embedded the line editor in, because the bottom portion of the screen gets wiped if the user moves the cursor while entering a command.

gwenn:
Indeed, it seems possible to move only the cursor like [here](https://github.com/nsf/termbox-go/blob/master/termbox.go#L83) or [here](https://github.com/nsf/termbox-go/blob/master/termbox_windows.go#L556).
I will try on unix.

gwenn:
Ok,
Moving only the cursor may be difficult in multi-line mode.
I am not sure but it seems to me that the culprit is the current strategy used to refresh the screen.
Because we are clearing from the prompt to the end of the screen (`\x1b[J`).
Could you please confirm that if we clear only the lines/rows drawn/used by the editor, your issue will be solved ?

dcampbell24:
@gwenn yes, I think that is the right strategy. Even better might be to take a rectangle or bounding box as a parameter. I also don't think that the prompt should write anything to output. This can be done by application code and separates concerns.

dcampbell24:
What I mean about writing to stdout is to not write the prefix or the line ending. That gives a lot more flexibility to the code consuming the library. For example if I want to have '\r\n' rather '\n' as the EOL.

gwenn:
Just so I fully understand, you want the following line to be removed:
https://github.com/kkawakam/rustyline/blob/master/src/lib.rs#L877
```rust
    println!("");
```
?

dcampbell24:
@gwenn yes, I think that line should either be removed, or the readline function should at least take an argument to allow for cross platform line endings. Thank you for your help.

gwenn:
Are you sure `println!("")` doesn't print `\r\n` on windows ?

dcampbell24:
> Are you sure println!("") doesn't print \r\n on windows ?

No.

I am more concerned with how it inter-operates with (termion)[https://github.com/ticki/termion]. I am going to open source the project I am using rustyline in, so I can give a link after I update the license. Also, I can open a new ticket to discuss line terminators if it makes sense to.

gwenn:
The associated patch has been merged in master.

dcampbell24:
Thanks!

PR #53, opened 2016-07-16
Clear only the lines/rows drawn/used by the editor (#48)

gwenn:
And do not ignorespace by default.
And do not panic when the history is empty.

### DIFF
--- a/src/history.rs
+++ b/src/history.rs
@@ -169,6 +169,7 @@ mod tests {
     #[test]
     fn add() {
         let mut history = super::History::new();
+        history.ignore_space(true);
         assert!(history.add("line1"));
         assert!(history.add("line2"));
         assert!(!history.add("line2"));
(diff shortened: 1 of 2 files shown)


src/history.rs
```diff
@@ -20,7 +20,7 @@ impl History {
         History {
             entries: VecDeque::new(),
             max_len: DEFAULT_HISTORY_MAX_LEN,
-            ignore_space: true,
+            ignore_space: false,
```
kkawakam: Whats the reason for this change?

gwenn: Because I checked and found no similar projet where lines which begin with a space character are ignored (`copperline`, `peterh/liner`, `haskeline`, `linenoise-ng`, `jline2`, ...).

kkawakam:
Closing since it was rolled up with #54

Issue #37, opened 2016-05-18
Support Windows platform


gwenn:
https://www.reddit.com/r/rust/comments/4jufvh/which_readlinelike_crate/d3afuv8
> I really wish there was a readline-like library which worked for windows as well.

kkawakam:
@gwenn I'm gonna start taking a look at this if you don't mind.  Going to refer to https://github.com/MSOpenTech/redis/tree/3.0/deps/linenoise which is the windows port of linenoise.

gwenn:
I don't have access to a Windows machine and I tried Windows VM without success.
And there is another fork with windows support here:
https://github.com/arangodb/linenoise-ng
Regards.

kkawakam:
Partial support added in #42.  Rust's stdio module for Windows uses the ReadConsoleW from winapi which does not seem to capture modifier keys (CTRL or ALT).  Some additional work needs to be done to instead use the ReadConsoleInputW in winapi to support the same operations as we do in Linux

Marwes:
@kkawakam History (pressing up arrow) does not work on Windows but I assume this is for the same reason as CTRL and ALT?

kkawakam:
@Marwes that is probably the case, you should probably continue using linenoise-rs for support on windows.  I am still working on full windows support right now and hopefully I'll have that out this week.

gwenn:
Ok,
Don't take me wrong but...
I started to look at Windows support by using [Wine](https://www.winehq.org) and cross compiling.
 * refresh screen:
https://github.com/gwenn/rustyline/blob/master/src/lib.rs#L192
 * clear screen:
https://github.com/gwenn/rustyline/blob/master/src/lib.rs#L416
 * console input:
https://github.com/gwenn/rustyline/blob/master/src/lib.rs#L959

I am also looking at [this](https://github.com/DanielKeep/rust-ansi-interpreter) project to support ANSI escape sequence on Windows.

If it looks good to you, I can take some time to merge and send a pull request.
Regards.

kkawakam:
@gwenn that is totally fine with me.  The only thing is that I would rather want to this platform specific stuff to live outside of lib.rs (kinda following the pattern we see in rust's std lib).  

Also can I add you as a collaborator, I don't have as much time to work on this at this moment and I don't want to really block your progress.

gwenn:
[...]

PR #54, opened 2016-07-25
Win32

gwenn:
As promised!
More coming...


gwenn:
Tested (not thoroughly) at work with cmd and powershell.
Fix #37

kkawakam:
I am good with all the changes you made. I would definitely would like to move the platform specific logic for clear_screen() and refresh() to the tty module (maybe not the best name for it?) but it doesn't have to be part of this PR.

gwenn:
I tried to move `clear_screen` but there is a visibility problem:
```
src/tty/unix.rs:106:29: 106:34 error: private type in public interface [E0446]
src/tty/unix.rs:106 pub fn clear_screen(s: &mut State) -> Result<()> {
```

kkawakam:
@gwenn, yea it isn't really important right now.  I'm sure something needs to be refactored to get that working.

kkawakam:
@gwenn Do you have more changes coming in?

gwenn:
I think so:
* Powershell & ctrl-L:
thread 'main' panicked at 'attempted to multiply with overflow', src\lib.rs:255
* Powershell ISE:
Error: Io(Error { repr: Os { code: 6, message: "The handle is invalid." } })

kkawakam:
@gwenn Out of the new issues you created, you're addressing  #63, #56, #55 specifically in this branch?  Just let me know when you're good to go, so I can take a final look + merge.

gwenn:
Definitely not #56.
Probably not #63.
Surely #55.

gwenn:
* For #63, could you please check if you receive WINDOW_BUFFER_SIZE_EVENT when resizing the console (with Cmd or Powershell) ? Just add a debug statement [here](https://github.com/gwenn/rustyline/blob/win32/src/tty/windows.rs#L135)
```
writeln!(io::stderr(), "rec: {:?}", rec).unwrap();
```
Normally, setting the ENABLE_WINDOW_INPUT [here](https://github.com/gwenn/rustyline/blob/win32/src/tty/windows.rs#L88)

 * For #56, it may be related to this one: https://github.com/retep998/wio-rs/issues/5. If linenoise-ng example works in Powershell ISE, there is a bug in rustyline.

kkawakam:
@gwenn for #63 Confirmed WINDOW_BUFFER_SIZE_EVENT when resizing the console in Cmd & Powershell

kkawakam:
So I was taking a look at #56 and it seems like interactive console applications aren't really supported in  Powershell ISE.
 * https://blogs.msdn.microsoft.com/powershell/2009/02/04/console-application-non-support-in-the-ise/

* https://blogs.msdn.microsoft.com/powershell/2009/04/17/differences-between-the-ise-and-powershell-console/

I also found this piece of code [link](https://github.com/neurocline/catalog/blob/master/cpp/src/WinConsoleHelpers.cpp#L32) that says:
```
// Grr, Powershell ISE says FILE_TYPE_PIPE all the time, evidently it runs as a pipe
// to the console? This might be an artifact of 2.0, evidently in 3.0 the output and
// command panes are merged.
```
This might be the reason why GetConsoleScreenBufferInfo is failing because stdout is a FILE_TYPE_PIPE and not a FILE_TYPE_CHAR.

Issue #57, opened 2016-07-27
Display all completion list


gwenn:
https://github.com/BurntSushi/tabwriter may be useful.

PR #74, opened 2016-09-04
Page completions

gwenn:
BREAKING CHANGE

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -2,7 +2,7 @@ extern crate rustyline;
 
 use rustyline::completion::FilenameCompleter;
 use rustyline::error::ReadlineError;
-use rustyline::Editor;
+use rustyline::{Config, Editor};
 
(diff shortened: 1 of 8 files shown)


gwenn:
Fix #57

kkawakam:
@gwenn sorry for the late response.  Can we limit the scope of the breaking API changes and just ```#[deprecated]``` the functions?  We can remove the deprecated functionality when we go 2.0.0 but I don't want it to prevent us from pushing minor version changes in the meanwhile.

gwenn:
Sorry but I don't really care.
Maybe we can go back to something like version 0.3 ?
Or leave this pull request unmerged until version 2.0 ?

kkawakam:
@gwenn can you rebase against master (or merge) ? I'm going to merge this in once that it is done. I think it's better to merge this in rather than waiting.  Thanks!

gwenn:
Done.

Issue #82, opened 2016-10-02
Add an Iterator


Yamakaky:
We could do:

``` rust
for cmd in editor {
    match ...
}
```

It would end on EOF or <C-d>

gwenn:
And for \<C-c\> ?

Yamakaky:
In [my code](https://github.com/Yamakaky/dcpu/blob/master/src/emulator/debugger/mod.rs), I ignore it. It's handy if you want to discard the current line. `python` does it.

gwenn:
Something like:
```rust
impl<C: Completer> Iterator for Editor<C> {
    type Item = Result<String>;

    fn next(&mut self) -> Option<Result<String>> {
        let readline = self.readline(""); // FIXME prompt
        match readline {
            l @ Ok(_) => Some(l),
            Err(error::ReadlineError::Eof) => None,
            e @ Err(_) => Some(e),
        }
    }
}
```
?

Yamakaky:
Yeah, something like that. The prompt could be configured at creation time, since it tend to stay the same for most program. You should also keep the current interface, in case someone want to do something different

Yamakaky:
SGTM. Just add an implementation for `IntoIterator` for extra convenience ^^

gwenn:
2.0 released

PR #84, opened 2016-10-08
Draft for #82

gwenn:


Yamakaky:
Why not just `impl Iterator for Rustyline`?

gwenn:
Because there is no default prompt.
Do you have any idea how to specify the prompt ?

Yamakaky:
Maybe just allow for a default prompt to be set?

kkawakam:
@Yamakaky @gwenn is this a satisfactory solution for this?  I am neutral on whether we specify a default prompt or not.

Yamakaky:
I'm not sure, either should be fine.

PR #95, opened 2016-11-03
fix autocomplete cancel

donhcd:
When you cycle through all the options and return to the original uncompleted input, we should update the State.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -550,7 +550,6 @@ fn complete_line<R: RawReader>(rdr: &mut R,
                 // Restore current edited line
                 s.snapshot();
                 try!(s.refresh_line());
-                s.snapshot();
             }
 
             key = try!(rdr.next_key(false));
@@ -561,12 +560,10 @@ fn complete_line<R: RawReader>(rdr: &mut R,
                         try!(beep());
                     }
                 }
-                KeyPress::Esc => {
+                KeyPress::Esc if i < candidates.len() => {
                     // Re-show original buffer
                     s.snapshot();
-                    if i < candidates.len() {
-                        try!(s.refresh_line());
-                    }
+                    try!(s.refresh_line());
                     return Ok(None);
                 }
                 _ => {


gwenn:
Hello,
Could you please explain why calling the completer is necessary ?
Is this related to a custom completer ?
Thanks.

donhcd:
you're right, I shouldn't need to call the completer. I was just using the default `completer::update` implementation so I can just inline these two lines instead:

```
        let end = line.pos();
        line.replace(start, end, elected)
```

donhcd:
okay, looks like this patch currently introduces another bug...

donhcd:
okay, I think I'm happy with this patch now

gwenn:
Which key do you use to exit/cancel completion ?
Do you reach this line https://github.com/kkawakam/rustyline/blob/master/src/lib.rs#L573 ?

donhcd:
The issue I'm having is when I cycle through all the options only pressing tab. Let's say I hardcode my completer to always return `(0, vec![String::from("bar"), String::from("baz")])`.

1) in the prompt, type: foo<TAB><TAB><TAB>
2) the prompt should show foo
3) hit enter, but "baz" is submitted

This makes sense looking at the code, because the two `snapshot()` calls make it so that the `line` not changed, even though the display shows a changed line

donhcd:
Yes, I do reach that line you linked but it `s.line()` is still holding the value of the last autocomplete that was cycled through

gwenn:
Ok, if you reach that line you are not **cancelling** the autocompletion but **validating** the autocompletion.
Currently, the only way to cancel the autocompletion is to hit the ESC key (see https://github.com/kkawakam/rustyline/blob/master/src/lib.rs#L570).
Maybe it is not the expected behaviour. I will check the other implementations.

donhcd:
@gwenn I don't see how the current behavior could be correct at all, and I don't understand why anybody would ever want a case where the line shows "foo" but the stored line is actually "bar". Let me followup with a gif of what's happening

donhcd:
I understand the idea that at that line you are validating the autocompletion, but the accepted autocompletion should be the original input which is being shown, not the last option that was suggested (and is no longer showing)

gwenn:
```
$ git diff
diff --git a/examples/example.rs b/examples/example.rs
index e87444d..a076561 100644
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -16,7 +16,7 @@ static PROMPT: &'static str = ">> ";
 fn main() {
         .history_ignore_space(true)
-        .completion_type(CompletionType::List)
+        //.completion_type(CompletionType::List)
         .build();
     let c = FilenameCompleter::new();
     let mut rl = Editor::new(config);
```
```
$ cargo run --example example
>> ls /usr/local/s<tab>
>> ls /usr/local/sbin/<tab>
>> ls /usr/local/share/<esc>
>> ls /usr/local/s
```
```
$ cargo run --example example
>> ls /usr/local/s<tab>
>> ls /usr/local/sbin/<tab>
>> ls /usr/local/share/<enter>
Line: ls /usr/local/share/
>> ls /usr/local/sh<tab>
>> ls /usr/local/share/<z><tab>
>> ls /usr/local/share/zsh/<esc>
>> ls /usr/local/share/z
```

donhcd:
here's the gif: http://www.giphy.com/gifs/26ybwEgsQ81wQYnRK

donhcd:
I'm not sure what you're saying in your last message... I'm using the `Default` config, so I'm already using `Circular`. In your first example, try hitting enter when you get to `ls /usr/local/s` the second time.

gwenn:
> In your first example, try hitting enter when you get to ls /usr/local/s the second time.

Indeed, the result is wrong.
I'm sorry I've taken so long to reproduce the problem.

gwenn:
It seems to me that the correct patch should be:
```
diff --git a/src/lib.rs b/src/lib.rs
index 9d978fc..5918c18 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -572,6 +572,9 @@ fn complete_line<R: RawReader>(rdr: &mut R,
                     return Ok(None);
                 }
                 _ => {
+                    if i == candidates.len() {
+                        s.snapshot();
+                    }
                     break;
                 }
             }
```

donhcd:
okay, I see that you're concerned about the ESC case, which I didn't consider. I've pushed an updated patch that feels a bit better to me

gwenn:
Maybe you should take a look at the [original linenoise code](https://github.com/antirez/linenoise/blob/master/linenoise.c#L387)...

donhcd:
I'm not sure what you're referring to... I see that there's a similar switch statement which involves esc and tab, but linenoise isn't some holy book that you're trying to translate, is it? If that was the goal, this repo could be much closer to this C code.

I don't see the purpose in convoluting this repo's code in order to emulate the AST of the another repo's code, which is written in a completely different language.

donhcd:
If the goal really is to emulate the linenoise code, then I'd think the first order of business would be to remove this `snapshot` method, which I don't see anywhere in that linenoise file

gwenn:
My point was that there is no condition on the ESC case (the key behaves the same while iterating over candidates).
And the `snapshot` is the equivalent to `struct linenoiseState saved = *ls;` except that I tried to avoid copying the whole state at each iteration step (which introduced the current bug)...
Lastly, linenoise is much more used than rustyline and as far as I know, no one complains about the ESC key behaviour of linenoise.

donhcd:
Yes, because C doesn't allow you to combine a switch case with a condition. The code you've added in #95 just seems to propagate the error of having two `snapshot` calls in the `Restore current edited line` case, which is plain convoluted.

gwenn:
Maybe an example is better:
```
$ git co https://github.com/antirez/linenoise
$ cd linenoise
$ make
$ ./linenoise_example
hello> h<tab>
hello> hello World<tab>
hello> hello there<tab>
hello> h<tab>
hello> hello World<esc>
hello> h
```
=> Good (the ESC key behaves the same no matter how many times the TAB key is pressed)

With your patch:
```
$ cargo run --example example
>> ls /usr/local/s<tab>
>> ls /usr/local/sbin/<tab>
>> ls /usr/local/share/<tab>
>> ls /usr/local/s<tab>
>> ls /usr/local/sbin/<esc>
>> ls /usr/local/share/
```
=> Wrong

With my patch:
```
$ cargo run --example example
>> ls /usr/local/s<tab>
>> ls /usr/local/sbin/<tab>
>> ls /usr/local/share/<tab>
>> ls /usr/local/s<tab>
>> ls /usr/local/sbin/<esc>
>> ls /usr/local/s<enter>
Line: ls /usr/local/s
```
=> Good

gwenn:
> which is plain convoluted.

I agree with that.
But I want to avoid copying the whole state at each iteration step.

(O) original line + cursor position (before completion)
(C1) updated line with first candidate + new cursor position
(C2) updated line with second candidate + new cursor position
...
The user presses TAB N times (N>0).
Then she presses ESC or another key (but not a TAB).
If ESC is pressed, (O) is restored.
If another key (but not a TAB) is pressed, the current candidate (which may be (0)) is accepted.

```rust
s.backup() // after this, State.snapshot contains (O)
loop {
  // Show completion or original buffer
  if i < candidates.len() {
      completer.update(&mut s.line, start, &candidates[i]); // after this, State.line containts (C1) or (C2)...
      try!(s.refresh_line());
  } else {
      // Restore current edited line
      s.snapshot(); // After this State.line contains (O) and State.snapshot contains (CN)
      try!(s.refresh_line());
      s.snapshot(); // After this State.snaphot contains (O)
  }

  key = try!(rdr.next_key(false));
  match key {
      KeyPress::Tab => {
          i = (i + 1) % (candidates.len() + 1); // Circular
          if i == candidates.len() {
              try!(beep());
          }
      }
      KeyPress::Esc => {
          // Re-show original buffer
          s.snapshot(); // After this State.line contains (O)
          if i < candidates.len() {
              try!(s.refresh_line());
          }
          return Ok(None);
      }
      _ => {
          if i == candidates.len() {
              s.snapshot(); // After this State.line contains (O)
          }
          break;
      }
  }
}
```

donhcd:
Okay, sorry for my misunderstanding. I didn't try going through the autocomplete options more than once, which makes the problem super obvious - the original state gets lost in my patch.

Thanks for explaining the issue (and for merging a fix to the problem I was concerned with)! I understand why the code is written this way now and I'm happy to close this pr

donhcd:
oops, I thought you'd committed that patch to this repo, but I see you made it to your fork. I've modified this patch to be what you suggested

gwenn:
Perfect!
Many thanks.

PR #96, opened 2016-11-05
Try to handle single ESC key

gwenn:
The side effect is that manual escape sequences such as ESC-f
must be typed quickly...


gwenn:
Please don't apply this patch.

PR #101, opened 2016-11-27
Introduce with_config construstor to ease migration

gwenn:
Only the projects using:
 - `History.search`
 - `History.ignore_space`
 - `History.ignore_dups`
 - `Editor.history_ignore_space`
 - `Editor.history_ignore_dups`
 - `Editor.set_history_max_len`
 - `History::default`
 - `Editor::default`

will now be impacted when upgrading from 1.0 to 2.0...

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -19,7 +19,7 @@ fn main() {
         .completion_type(CompletionType::List)
         .build();
     let c = FilenameCompleter::new();
-    let mut rl = Editor::new(config);
+    let mut rl = Editor::with_config(config);
     rl.set_completer(Some(c));
     if rl.load_history("history.txt").is_err() {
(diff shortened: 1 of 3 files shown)


gwenn:
Impacted:
 * https://github.com/sector-f/unnamed-shell/blob/6a5a536b4c80193fce7c4c429cecfb2f28cede8d/src/bin/rustyline-shell.rs#L13
 * https://github.com/matthiasbeyer/imag/blob/ac1af164e902b085895d76b8163f9fa41df4e3d1/libimaginteraction/src/readline.rs#L86
 * https://github.com/turboladen/ruosh/blob/6e77e84926898d863e1c6b428ebc1916c57657ea/src/main.rs
 *

Issue #94, opened 2016-10-30
vi mode


thiolliere:
Have you consider implement vi mode ?

however great project thanks.

gwenn:
Maybe we should try to port  [this](https://github.com/antirez/linenoise/pull/92) patch ?

gwenn:
2.0 released

PR #103, opened 2016-12-17
Vi mode (#94)

gwenn:
Still missing:
 - In Vi command mode, prevent user from going to end of line.

### DIFF
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -45,6 +45,16 @@ pub enum Cmd {
 }
 
 impl Cmd {
+    pub fn should_reset_kill_ring(&self) -> bool {
+        match *self {
+            Cmd::Kill(Movement::BackwardChar(_)) |
+            Cmd::Kill(Movement::ForwardChar(_)) => true,
+            Cmd::ClearScreen | Cmd::Kill(_) | Cmd::Noop | Cmd::Suspend | Cmd::Yank(_, _) |
+            Cmd::YankPop => false,
+            _ => true,
+        }
+    }
+
     fn is_repeatable_change(&self) -> bool {
         match *self {
             Cmd::Insert(_, _) => true,
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -86,7 +86,7 @@ pub struct EditState {
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum Movement {
-    WholeLine,
+    WholeLine, // not really a movement
     BeginningOfLine,
     EndOfLine,
     BackwardWord(RepeatCount, Word),
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -110,11 +110,12 @@ impl EditState {
         }
     }
 
-    fn digit_argument<R: RawReader>(&mut self,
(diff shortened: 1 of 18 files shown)


kkawakam:
Sorry was on holiday, I'm taking a look at this now.

src/keymap.rs
```diff
@@ -0,0 +1,416 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Copy)]
+pub enum At {
+    Start,
+    End,
+}
+
+#[derive(Debug, Clone, PartialEq, Copy)]
+pub enum Anchor {
```
kkawakam: I know it is used for yank, but a comment would help here.

src/keymap.rs
```diff
@@ -0,0 +1,416 @@
...
+    ViCharSearch(i32, CharSearch),
+    ViDeleteTo(i32, CharSearch),
+    Yank(i32, Anchor),
+    YankPop,
+}
+
+#[derive(Debug, Clone, PartialEq, Copy)]
+pub enum Word {
+    // non-blanks characters
+    Big,
```
gwenn: It is the term used at least by libedit, haskeline:
> A "bigword" consists of: a maximal sequence of non-blanks preceded and followed by blanks

src/keymap.rs
gwenn: I chose this word because 'where' is a Rust keyword...
Any suggestion is welcome.

kkawakam: Ah thats fine then.

src/keymap.rs
kkawakam: Oh okay.  Can you put that in as the comment?

gwenn:
There is [one commit](https://github.com/gwenn/rustyline/commit/454cf36fde7e69e50658f99566cd6aa322df152d) that I cannot add to this PR because it needs the `unicode` feature:
https://doc.rust-lang.org/nightly/std/primitive.str.html#method.is_alphanumeric

src/keymap.rs
```diff
@@ -0,0 +1,527 @@
...
+                                  rdr: &mut R,
+                                  config: &Config,
+                                  digit: char)
+                                  -> Result<KeyPress> {
+        self.num_args = digit.to_digit(10).unwrap() as i16;
+        loop {
+            let key = try!(rdr.next_key(config.keyseq_timeout()));
+            match key {
+                KeyPress::Char(digit @ '0'...'9') => {
+                    self.num_args = self.num_args * 10 + digit.to_digit(10).unwrap() as i16;
```
iamcodemaker: This can overflow, might want to use `wrapping_add()`/`wrapping_mul()` or even better `checked_add()`/`checked_mul()`. In debug mode overflow would cause a panic. In release mode, the count would simply be wrong. Though in the event of an overflow, the count is going to be wrong anyway no matter what you do, so overall this is a minor issue.

gwenn: Thanks.
I will try to fix this.

iamcodemaker:
`sautrating_mul()`, even better!

PR #105, opened 2017-01-05
Make completion::ESCAPE_CHAR public

klemens:
This is necessary to correctly use the `unescape` function on every platform (i.e. to extract the path after it has been completed).


klemens:
Closed in favour of a proper solution, see #106.

PR #114, opened 2017-02-19
Adds support for "blinking" matching delimiters `({[`

### LINKED ISSUE #111: Add support for matching delimiter blinking
I would like rustyline to blink the matching delimiter once I type it.  Is there any reason you see why this shouldn't work in the library? Otherwise, I can put together a pull request.

I would think to start with `()` but I don't see an immediate reason to not add `{} []`.  Perhaps it should be configurable for the user?

ralexstokes:
Briefly moves the cursor to a matching delimiter when one is typed, then returns the cursor to its original position either after a configurable timeout or the user supplies more input.

Resolves #111

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -13,6 +13,8 @@ pub struct Config {
     completion_prompt_limit: usize,
     /// Duration (milliseconds) Rustyline will wait for a character when reading an ambiguous key sequence.
     keyseq_timeout: i32,
+    /// Maximum duration (milliseconds) Rustyline will use for animating the blink on a matching delimiter.
+    matching_delimiter_timeout: i32,
 }
 
 impl Config {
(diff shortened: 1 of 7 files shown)


gwenn:
Thanks for your PR.

The Windows implementation of `next_char_ready` is missing.
And the associated declaration must not be tagged with `#[cfg(unix)]`.
[Linefeed](https://github.com/murarth/linefeed/blob/master/src/reader.rs#L839) uses `WaitForSingleObject` to implement `next_char_ready` on Windows.

Alternatively, [it](https://github.com/arangodb/linenoise-ng/blob/master/src/linenoise.cpp#L1192) seems possible to highlight matching brace without depending on `next_char_ready`.
Regards.

ralexstokes:
Thanks for the feedback!  I have added Windows support via `WaitForSingleObject`.  I looked at `linenoise-ng` but when I ran the example, it did not seem to have this feature.

CI is passing although I do not have a Windows machine handy to try it out.

gwenn:
Indeed, `linenoise-ng` has a different behaviour: matching brace is highlighted only when the cursor is on brace (not when a brace is inserted).

gwenn:
See #163

gwenn:

PR #118, opened 2017-05-15
Add support for Ctrl-Left and Ctrl-Right arrow key combinations

fintelia:


gwenn:
There is a pending [PR](#103) which introduces an indirection (a keymap) between the key press and the action/command.
And there is one [commit](https://github.com/gwenn/rustyline/commit/f1bb612b16a5d18b780583dd1dd6bb0b04f77583) which add Ctrl-Left and Ctrl-Right for both unix and windows but these keys are not bound by default...

gwenn:

PR #120, opened 2017-07-06
Add support for surrounding the prompt by editable characters.

BourgondAries:
This feature is used to implement a lisp-like line reader, where
surrounding the cursor by '(|> ' and ')' is used to make the shell
more convenient for the user to use. (`|>` can be a macro that finds executables 
or does something useful in lisp).

The method by which the cursor moves backward seems rather clumsy, I'd rather move the cursor back by N utf-8 characters instead of looping. Even better would be to use a method to insert after the cursor whilst keeping the cursor in its current position.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1018,6 +1018,265 @@ fn readline_edit<C: Completer>(prompt: &str,
     Ok(s.line.into_string())
 }
 
+/// Handles reading and editting the readline buffer.
+/// It will also handle special inputs in an appropriate fashion
+/// (e.g., C-c will exit readline)
+#[allow(let_unit_value)]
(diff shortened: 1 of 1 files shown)


gwenn:
You want something like [this](http://hackage.haskell.org/package/haskeline-0.7.4.0/docs/System-Console-Haskeline.html#v:getInputLineWithInitial), no ?

gwenn:
It is not as flexible as haskeline but [liner](https://github.com/MovingtoMars/liner/commit/5862d60d8ef477f4b0ee9163906b4f4c501afa53) supports something similar.

gwenn:
See https://github.com/kkawakam/rustyline/commit/e949ef98f7176dc43ee27a6980a8883799358a0c

PR #121, opened 2017-07-15
Add support for using Ctrl/Alt/Meta with non-character keys

dubrowgn:
Currently, key modifiers can only be used in conjunction with keys which map to a printable character. This pull request is a refactor that allows key modifiers to be used with any key, such as arrow keys, tab, etc.

### DIFF
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -1,68 +1,101 @@
-
 #[derive(Debug, Clone, PartialEq, Copy)]
-pub enum KeyPress {
-    UnknownEscSeq,
+pub enum Key {
     Backspace,
     Char(char),
-    Ctrl(char),
     Delete,
     Down,
     End,
     Enter, // Ctrl('M')
     Esc,
     Home,
+    Insert,
     Left,
-    Meta(char),
     Null,
     PageDown,
     PageUp,
     Right,
     Tab, // Ctrl('I')
+    Unknown,
     Up,
+}
+
+#[derive(Debug, Clone, PartialEq, Copy)]
+pub struct KeyPress {
+    pub key: Key,
+    pub alt: bool,
+    pub ctrl: bool,
+    pub shift: bool,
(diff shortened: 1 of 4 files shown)


dubrowgn:
It's a good point. What do you think of something like this then?

```rust
pub enum Key {
    Backspace,
    Char(char),
    Delete,
    Down,
    End,
    Enter,
    Esc,
    Home,
    Insert,
    Left,
    Null,
    PageDown,
    PageUp,
    Right,
    Tab,
    Unknown, // not sure about this, from UnknownEscSeq
    Up,
}

pub struct KeyPress {
    pub key: Key,
    pub ctrl: bool,
    pub alt: bool,
    pub shift: bool,
    pub meta: bool,
}
```

gwenn:
`alt`and `meta` are the same by default on most unix keyboard, no ?
https://unix.stackexchange.com/questions/28993/what-is-bashs-meta-key:
> By default, on most setups, the Meta key is the key labeled Alt. This is because historically, many unix workstations had a key labeled Meta where PCs have a key labeled Alt.

dubrowgn:
In many other contexts the meta key is the same as the super key (aka windows, command, etc).

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey
https://forum.kde.org/viewtopic.php?f=111&t=88599

For clarity, I suppose it would be better to call it super here, instead of meta.

dubrowgn:
So, I'm playing around with this a bit. I ended up implementing the above proposal and several macro helpers. What do you think?

gwenn:
I don't know if the pending PRs will be integrated one day and, if they are, in which order.
But your PR and #103 will generate a lot of conflicts.

dubrowgn:
Indeed, although I suspect this PR will cause many conflicts with just about any other change. There is a lot being touched for this.

gwenn:

PR #130, opened 2018-05-08
Add Fuchsia OS Support

cavedweller:


src/tty/fuchsia.rs
```diff
@@ -0,0 +1,244 @@
...
+
+// TODO properly set raw mode, process escape keys
+
+impl RawReader for FuchsiaRawReader {
+    fn next_key(&mut self) -> Result<KeyPress> {
+        let c = try!(self.next_char());
+
+        let mut key = consts::char_to_key_press(c);
+        if key == KeyPress::Esc {
+            // TODO
```
cavedweller: pollfd isn't quite as simple on Fuchsia and I haven't gotten around to it yet.

Cargo.toml
```diff
@@ -21,9 +16,13 @@ unicode-width = "0.1.3"
 unicode-segmentation = "1.0"
 encode_unicode = "0.1.3"
 
-[target.'cfg(unix)'.dependencies]
+[target.'cfg(all(unix, not(any(target_os = "fuchsia"))))'.dependencies]
 nix = "0.8"
 
+[target.'cfg(target_os = "fuchsia")'.dependencies]
+fuchsia-zircon = "0.3.2"
+fuchsia-device = "0.1.0"
```
cavedweller: This crate still needs to be published on crates.io

src/tty/fuchsia.rs
gwenn: Most of the time a single ESC is not expected (at least in Emacs mode (except after Ctrl-R)).
So I suggest you try reading an escape sequence.
User can just press ESC twice as a workaround...

Cargo.toml
gwenn: Thanks you for this PR.
I will give your PR a try. I just need to figure how to install Fuchsia...
And I must warn you.
Currently, `rustyline` is stalled (but not dead).
But there are, at least, two alternatives: [liner](https://github.com/MovingtoMars/liner) and [linefeed](https://github.com/murarth/linefeed).
So maybe, you should consider these librairies as better alternatives...

src/tty/fuchsia.rs
```diff
@@ -0,0 +1,244 @@
...
+}
+
+impl FuchsiaRawReader {
+    pub fn new() -> Result<FuchsiaRawReader> {
+        let stdin = StdinRaw {};
+        Ok(FuchsiaRawReader {
+            chars: char_iter::chars(stdin),
+        })
+    }
+    fn escape_sequence(&mut self) -> Result<KeyPress> {
```
gwenn: Do you know if escape sequences are VT100 ?
If yes, we should try to factorize `escape_sequence` between all unix platforms...

Cargo.toml
cavedweller: linefeed had too much GNU/Posix-isms in it for my taste. I'll check out liner though!

I'd recommend this resource: 
https://fuchsia.googlesource.com/docs/+/master/getting_started.md
If you have any problems trying Fuchsia, shoot me an email at bwb@google.com.

src/tty/fuchsia.rs
cavedweller: Yup! It's at least pretty close to vt100

cavedweller: ... as most things are :P

Cargo.toml
gwenn: Sorry, I gave up installing Fucshia: I have only an old MacBook Air...
Could please confirm that there is no pre-built vbox ?
Thanks.

PR #133, opened 2018-06-09
Add a way to get the helper

### LINKED ISSUE #116: Add a way to retrieve the current completer
I was messing with rustyline, it's completion API and trying do implement multiline editing at the same time, but then I encountered a problem: I need to have the input from the previous lines to have a proper context for my completion system.

Sorry if it's unclear, here's a quick example:

```
input> if [ x -eq 1 ];
  ...> th<tab>
```

In this case, my completer need to know that there's an `if` before to be able to complete the `then`.

My idea was to store the previous lines as context inside the Completer, but currently there is no way of retrieving the completer in use.

A see two ways of doing this: either make the `completer` field public or add the `get_completer` and `get_completer_mut` methods.

For now, I can do it by replacing the current completer with the added context, but I think that would be better by mutating the completer directly.

If there's a better way, please tell me, I'm open to discussion ;-)

robalni:
This fixes #116 

I have a similar problem to the one described in that issue in one of my projects so I'm fixing it now.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -753,7 +753,7 @@ impl<H: Helper> Editor<H> {
         self.helper.as_mut()
     }
 
-    /// Return an imutable reference to the helper.
+    /// Return an immutable reference to the helper.
     pub fn get_helper_const(&self) -> Option<&H> {
         self.helper.as_ref()
     }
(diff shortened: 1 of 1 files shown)


gwenn:
Thanks.

PR #143, opened 2018-08-28
Use into() instead of casting to c_ulong because of different types in some targets

mssun:
For example, in `x86_64-unknown-linux-musl`, `ioctl` takes `c_int`. It is better to use `into()` to work around this issue than using `as c_ulong`.

https://rust-lang.github.io/libc/x86_64-unknown-linux-musl/libc/fn.ioctl.html
```
pub unsafe extern "C" fn ioctl(fd: c_int, request: c_int, ...) -> c_int
```

### DIFF
--- a/src/tty/unix.rs
+++ b/src/tty/unix.rs
@@ -33,7 +33,8 @@ fn get_win_size() -> (usize, usize) {
     unsafe {
         let mut size: libc::winsize = zeroed();
         // https://github.com/rust-lang/libc/pull/704
-        match libc::ioctl(STDOUT_FILENO, libc::TIOCGWINSZ as libc::c_ulong, &mut size) {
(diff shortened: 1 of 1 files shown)


gwenn:
I was using `into()` but a warning was generated by clippy or/and rustc on macos and/or linux platform so I made this change:
https://github.com/kkawakam/rustyline/commit/cb056d3a344f5a5642c7fc3932abb65337db9368#diff-0655ced7de6892bafa1f0000619c718eR35
So I am not sure that
> It is better to use into() to work around this issue than using as c_ulong
...

mssun:
Currently, it fails to compile on `musl`. Therefore, at least it's better than warning because of a bug in libc.

gwenn:
Thanks.
I will fix the warning with an `#[allow(???)]`...

PR #151, opened 2018-09-14
Ignore all IO errors during completion

eminence:
Previously, there were two ways to have an error thrown during completion:

* try to tab complete a directory that didn't exist at all (example `/blah/`<kbd>tab</kbd> where `/blah/` doesn't exist).
* try to tab complete in a directory without permissions (example: create `/tmp/foo` and then `chmod 000 /tmp/foo` and then in rustyline `/tmp/foo/`<kbd>tab</kbd>).

This change ignores all errors when doing tab completion

### DIFF
--- a/src/completion.rs
+++ b/src/completion.rs
@@ -299,20 +299,30 @@ fn filename_complete(
     };
 
     let mut entries: Vec<Pair> = Vec::new();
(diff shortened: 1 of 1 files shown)


gwenn:
Thanks.

PR #158, opened 2018-10-19
Complete hints with the right key

Flakebi:
Hi,
I did not find a way to complete the displayed hint, so I added one.
Pressing the `right` key, when the cursor is at the end if the input, i.e. just before the hint, inserts the hint.
If the cursor is not at the end of the line or no hint is available, the default command is executed (move one to the right).


gwenn:
First, thank you for your PR.
I suggest adding `is_cursor_at_end` and `has_hint` to the `Refresher` trait.
And use them to choose between `Cmd::Move(...)` or `Cmd::CompleteHint` so you will not have to box anything.

Flakebi:
Good idea, that’s a lot better than with Box.

Flakebi:
Done 🙂

PR #169, opened 2018-11-25
Add disable_flush_newline

riywo:
This commit is for adding an optional feature called `disable_flush_newline`. By default, it is `false` so that the current behaviour has no change.

## Object
To give option when a developer wants to add something after user's input before newline. For example, [gforth](https://www.forth.com/starting-forth/1-forth-stacks-dictionary/) does:

```
> 1 . [ENTER] 1  ok
> 
```

## Feature
If a developer specify `disable_flush_newline` as `true`, newline won't output after user's input. The output using the README example is like below:

### `true`
```
>> aaaaLine: aaaa
>> 
```
### `false` (default behaviour)
```
>> aaaa
Line: aaaa
>> 
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -24,6 +24,9 @@ pub struct Config {
     color_mode: ColorMode,
     /// Whether to use stdout or stderr
     output_stream: OutputStreamType,
+    /// If true, newline won't be output when a user input Enter.
+    /// This is useful when a developer wants to output something after user's input.
+    disable_flush_newline: bool,
 }
 
 impl Config {
@@ -109,6 +112,8 @@ impl Config {
     pub(crate) fn set_output_stream(&mut self, stream: OutputStreamType) {
         self.output_stream = stream;
     }
+
+    pub fn disable_flush_newline(&self) -> bool { self.disable_flush_newline }
 }
 
 impl Default for Config {
@@ -124,6 +129,7 @@ impl Default for Config {
             auto_add_history: false,
             color_mode: ColorMode::Enabled,
             output_stream: OutputStreamType::Stdout,
+            disable_flush_newline: false,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


gwenn:
Sorry for the late reply.
Could you please show me `gforth` (or any readline-like library) code that implements this behaviour ?

riywo:
I don't know how gforth terminal works exactly, but this code might be related: https://github.com/forthy42/gforth/blob/master/engine/io.c

gwenn:
Could you please give #231 a try ?
For example,
With this PR and this simple patch:
```diff
Index: examples/example.rs
===================================================================
--- examples/example.rs	(revision 8c298f2ef6289ff8ae3b4a7e08bf8e6d4b1975e2)
+++ examples/example.rs	(date 1558269613000)
@@ -6,7 +6,8 @@
 use rustyline::error::ReadlineError;
 use rustyline::highlight::{Highlighter, MatchingBracketHighlighter};
 use rustyline::hint::{Hinter, HistoryHinter};
-use rustyline::validate::Validator;
+use rustyline::line_buffer::LineBuffer;
+use rustyline::validate::{Validator, ValidationResult};
 use rustyline::{Cmd, CompletionType, Config, Context, EditMode, Editor, Helper, KeyPress};
 
 static COLORED_PROMPT: &'static str = "\x1b[1;32m>>\x1b[0m ";
@@ -56,7 +57,12 @@
     }
 }
 
-impl Validator for MyHelper {}
+impl Validator for MyHelper {
+    fn validate(&self, line: &mut LineBuffer) -> ValidationResult {
+        let msg = format!(" . [ENTER] {} ok", line.as_str());
+        ValidationResult::Valid(Some(msg))
+    }
+}
 
 impl Helper for MyHelper {}
```
I can do something like this:
```sh
>> 1 . [ENTER] 1 ok
Line: 1
>> 
```

gwenn:
Should be fixed by #231

PR #172, opened 2018-12-06
Fix spelling in documentation

wirelyre:
I noticed some misspelled words in doc comments, so I ran `aspell` on the whole code base.

There are some more minor things that would be nice, like capitalizing "Unix" throughout, but I left these alone because it would be an annoying diff.

The README is a little cleaner now. Project names in the comparison table are capitalized according to how their respective authors do it. Just a little respectful gesture. Also, they are sorted alphabetically.


gwenn:
Thanks.

PR #181, opened 2018-12-28
Explain multiple line support better

RadicalZephyr:
It seems like the limited information about multi-line support in the readme causes wishful thinking to kick in and make people believe that the feature is more complicated than it really is.

Hopefully this should reduce that!


gwenn:
Thanks.

PR #182, opened 2018-12-29
Add multi-line editing support

RadicalZephyr:
This is my first draft on this idea.

I think it's pretty solid and the tests still pass.  I haven't tested it on any other platforms but Linux yet.

I'm definitely interested if anyone has feedback on the design of the trait.  I kept it really simple initially, and it suits my usage.  I think it should be flexible enough to be generally useful.

I wouldn't merge this until someone tests it on Mac and Windows.

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -56,11 +56,7 @@ impl Highlighter for MyHelper {
     }
 }
 
-impl Validator for MyHelper {
-    fn is_valid(&self, _line: &str) -> bool {
-        true
-    }
-}
+impl Validator for MyHelper {}
 
 impl Helper for MyHelper {}
 
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -6,6 +6,7 @@ use rustyline::config::OutputStreamType;
 use rustyline::error::ReadlineError;
 use rustyline::highlight::{Highlighter, MatchingBracketHighlighter};
 use rustyline::hint::{Hinter, HistoryHinter};
+use rustyline::validate::Validator;
 use rustyline::{Cmd, CompletionType, Config, Context, EditMode, Editor, Helper, KeyPress};
 
 static COLORED_PROMPT: &'static str = "\x1b[1;32m>>\x1b[0m ";
@@ -52,6 +53,12 @@ impl Highlighter for MyHelper {
 
     fn highlight_char(&self, line: &str, pos: usize) -> bool {
         self.1.highlight_char(line, pos)
+    }
+}
+
+impl Validator for MyHelper {
+    fn is_valid(&self, _line: &str) -> bool {
+        true
     }
 }
 
--- a/src/edit.rs
+++ b/src/edit.rs
@@ -14,6 +14,7 @@ use crate::keymap::{InputState, Refresher};
 use crate::line_buffer::{LineBuffer, WordAction, MAX_LINE};
 use crate::tty::{Position, RawReader, Renderer};
 use crate::undo::Changeset;
+use crate::validate::ValidationResult;
 
 /// Represent the state during line editing.
 /// Implement rendering.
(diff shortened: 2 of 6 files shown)


hayd:
What would this do with the history? Will it save the multi-line block as one entry?

When you say "multi-line editing", does this mean you can press up and edit the line above?

RadicalZephyr:
@hayd Yes, this saves the multiple line block as a single entry.  

Essentially, this change just lets the user insert literal newline characters into the buffer if the `Helper` judges that current buffer is "invalid" (i.e. incomplete).

I didn't implement navigation across buffer lines using the up or down motion commands, but you can use the forward/back motion commands to navigate across the buffer text as normal, and this includes crossing newlines in the buffer. This is a bit less convenient for use, but I think the increased complexity in overloading the up/down motion commands is not really worth it.

etaoins:
I tested this with my compiler's REPL in etaoins/arret#9. It looks very promising!

It works great in isolation but I get a weird interaction with hints. When I type the first line without pressing enter it looks like this:

intital

Note the red `)` indicating an unclosed bracket. If I then press enter the hint moves to the next line:
press-enter

If I press another key it then duplicates the prompt:
type-text-on-new-line

This is reproducible using Terminal.app both with macOS and SSHed in to Ubuntu

gwenn:
@etaoins I am not sure but I think that the bug is not related/introduced by this PR.
Could you please try to do the following with the `rustyline = "3.0.0"` on an unix platform (quoted-insert does not work on windows):
everything after ` # ` is what should be typed...
$ cargo run --example example
>> line1 # type 'l' 'i' 'n' 'e' '1' Ctrl-v Ctrl-j
line2 # type 'l' 'i' 'n' 'e' '2', now if delete key is pressed the prompt is duplicated.

The `PosixRenderer` is buggy...
I will try to fix this.

etaoins:
@gwenn You're correct, I get the same behaviour with rustyline 3.0.0 on macOS

src/validate.rs
```diff
@@ -0,0 +1,24 @@
...
+    /// indicating whether it is valid or not. The most common
+    /// validity check to implement is probably whether the input is
+    /// complete or not, for instance ensuring that all delimiters are
+    /// fully balanced.
+    ///
+    /// If you implement more complex validation checks it's probably
+    /// a good idea to also implement a `Hinter` to provide feedback
+    /// about what is invalid.
+    #[allow(unused_variables)]
+    fn is_valid(&self, line: &str) -> bool {
```
etaoins: I wonder if `is_valid` is the best name here. There could be cases where the line is syntactically invalid in a way that appending a more characters couldn’t fix. For example, this is terminally broken because the braces don’t match:

```rust
fn my_function(foo: i64]
```

Maybe `is_complete`, `is_closed`, `has_open_delimiter` could be better options?

examples/example.rs
```diff
@@ -55,6 +56,12 @@ impl Highlighter for MyHelper {
     }
 }
 
+impl Validator for MyHelper {
```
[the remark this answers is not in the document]
RadicalZephyr: true.  Originally there was no default implementation, forgot to remove this after I added it.

src/edit.rs
```diff
@@ -80,6 +84,13 @@ impl<'out, 'prompt> State<'out, 'prompt> {
             if let Ok(Cmd::Replace(_, _)) = rc {
                 self.changes.borrow_mut().begin();
             }
+            if let Some(validator) = self.validator {
```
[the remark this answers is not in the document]
RadicalZephyr: Do you mean just the check for whether the validator is present? If so, I'm not sure how that would work.

If you mean move this whole block, then I'll need to... completely change the implementation I guess. :)

src/validate.rs
RadicalZephyr: I'm also dissatisfied with the name.  I wanted to stay away from "complete" language since there is already a `Completer` helper trait with a `complete` function on it that is completely different in functionality.  The other names seem rather narrowly focused on the cases of closing delimiters but there's no real reason why that is the only usage for this.  Though it's probably a major one.

etaoins: How about something direct like `should_continue_line`?

daynin:
I think it will be better to add Shift+Enter support. So "Shift+Enter" means "next line" in this case and "Enter" means "end of input". It can be an optional parameter as well

hayd:
Supporting both would be awesome, but i don't think it all needs to be in this PR...

chris-henderson-alation:
A comment from the peanut gallery - I'm looking at this with keen interest as I am implementing a REPL frontend to a backend JDBC server, and having this would be a killer for multline queries.

hayd:
It looks like this work has been moved to #182!

PR #186, opened 2019-01-13
Fix bug in `HistoryHinter` when `ctx.history_index()` is 0

Ryan1729:
Running the example without a `history.txt` file on the current master lead to the following error: 
```
thread 'main' panicked at 'attempt to subtract with overflow', src\hint.rs:26:13 note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace. error: process didn't exit successfully: `target\debug\examples\example.exe` (exit code: 101)
```


gwenn:
Thanks.

PR #195, opened 2019-01-26
feature toggle dirs to legacy support

rustysec:
added a quick feature toggle for the [dirs](https://github.com/soc/dirs-rs) dependency. dirs is not legacy os (xp, server 2003, etc) compatible so it breaks any executable run on those platforms. the only feature `rustyline` loses is home directory completion, which i feel is a fine trade off.

### DIFF
--- a/src/completion.rs
+++ b/src/completion.rs
@@ -282,10 +282,8 @@ fn filename_complete(
     break_chars: &[u8],
     quote: Quote,
 ) -> Result<Vec<Pair>> {
-    #[cfg(feature = "dirs")]
-    {
-        use dirs::home_dir;
-    }
+    #[cfg(feature = "with-dirs")]
+    use dirs::home_dir;
     use std::env::current_dir;
 
     let sep = path::MAIN_SEPARATOR;
@@ -308,7 +306,7 @@ fn filename_complete(
                 dir_path.to_path_buf()
             }
         }
-        #[cfg(not(feature = "dir"))]
+        #[cfg(not(feature = "with-dirs"))]
         {
             dir_path.to_path_buf()
         }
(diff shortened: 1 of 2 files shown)


src/completion.rs
```diff
@@ -282,7 +282,10 @@ fn filename_complete(
     break_chars: &[u8],
     quote: Quote,
 ) -> Result<Vec<Pair>> {
-    use dirs::home_dir;
+    #[cfg(feature = "dirs")]
```
gwenn: ` #[cfg(feature = "with-dirs")]`

src/completion.rs
```diff
@@ -282,7 +282,10 @@ fn filename_complete(
...
+    {
```
gwenn: `{`, `}` can be removed.

src/completion.rs
```diff
@@ -294,12 +297,19 @@ fn filename_complete(
...
+                match dir_path.strip_prefix("~") {
+                    Ok(rel_path) => home.join(rel_path),
+                    _ => home,
+                }
+            } else {
+                dir_path.to_path_buf()
             }
-        } else {
+        }
+        #[cfg(not(feature = "dir"))]
```
gwenn: `#[cfg(not(feature = "with-dirs"))]`

src/completion.rs
rustysec: fixed

src/completion.rs
rustysec: fixed!

src/completion.rs
rustysec: got bit by last minute find and replaces... sorry about that.

gwenn:
Thanks.

PR #198, opened 2019-01-28
Update nix dependency to 0.13

silwol:
This is a semver breaking change, because the nix::Error type is exposed in the publicly exposed rustyline::ReadlineError enum.


gwenn:
Thanks.
Not the first and not the last breaking change...

PR #200, opened 2019-01-31
WIP: async start

RadicalZephyr:
This is start on an example demonstrating how to use rustyline from an async context.


remexre:
What all needs to be done here? I'd really like to get my hands on this :)

Issue #217, opened 2019-05-17
Compilation fails on nightly-2019-05-16


lopopolo:
Compilation is broken on the latest nightly due to a change in type inference:

```text
   Compiling rustyline v4.0.0
error[E0283]: type annotations required: cannot resolve `std::string::String: std::convert::AsRef<_>`
   --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/rustyline-4.0.0/src/lib.rs:627:43
    |
627 |             editor.add_history_entry(line.as_ref());
    |                                           ^^^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0283`.
error: Could not compile `rustyline`.
warning: build failed, waiting for other jobs to finish...
error: build failed
Exited with code 101
```

rustc version:

```text
$ rustc --version --verbose
rustc 1.36.0-nightly (7d5aa4332 2019-05-16)
binary: rustc
commit-hash: 7d5aa43325ad7629766b1183011f5bf5b2a1ea26
commit-date: 2019-05-16
host: x86_64-unknown-linux-gnu
release: 1.36.0-nightly
LLVM version: 8.0
$ cargo --version --verbose
cargo 1.36.0-nightly (c4fcfb725 2019-05-15)
release: 1.36.0
commit-hash: c4fcfb725b4be00c72eb9cf30c7d8b095577c280
commit-date: 2019-05-15
```

Example failed build: https://circleci.com/gh/lopopolo/ferrocarril/78

gwenn:
We need help to make #218 works on stable and nightly.

gwenn:
Should be fixed by release 4.1

PR #218, opened 2019-05-17
Try to fix nightly build

gwenn:
See #217

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -80,8 +80,8 @@ fn main() {
     loop {
         let readline = rl.readline(PROMPT);
         match readline {
-            Ok(ref line) => {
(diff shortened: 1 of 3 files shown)


gwenn:
We need help to make the fix works on stable and nightly.

nevi-me:
What's the earliest nightly that this breaks on? I'm trying to reproduce on Windows as it's breaking our builds

lopopolo:
@nevi-me somewhere between `nightly-2019-05-07` and `nightly-2019-05-16`

nevi-me:
I might have a fix, will submit a PR shortly. It's not 'nice' though

nevi-me:
I've submitted #219

PR #224, opened 2019-05-18
Make sure prompt is at the leftmost edge of the screen

### LINKED ISSUE #216: The last line of text printed by the program disappears if it doesn't end with a newline
Sample program:

```rust
use rustyline::{Editor,error::ReadlineError};

fn main() {
    let mut e: Editor<()> = Editor::new();

    loop {
        let rl = e.readline("$ ");
        match rl {
            Ok(_) => {
                print!("text\nwithout\nnewline");
            }
            Err(ReadlineError::Interrupted) | Err(ReadlineError::Eof) => break,
            Err(err) => panic!(err),
        }
    }
}
```

Expected output:

```
$ some text here
text
without
newline$ some other
```

Actual output:

```
$ some text here
text
without
$ some other
```

gwenn:
Fix #216

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -8,7 +8,9 @@ Bell
 
 Color
 - [X] ANSI Colors & Windows 10+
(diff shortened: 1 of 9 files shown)


gwenn:
Does not work on windows (tested on Windows 7) !

Issue #193, opened 2019-01-22
Rendering bug when a linefeed is inserted at the end of the edited line 


gwenn:
Reported [here](https://github.com/kkawakam/rustyline/pull/182#issuecomment-455091279).
The bug is related to [this](https://github.com/kkawakam/rustyline/blob/master/src/edit.rs#L246) code/optimisation.

gwenn:
Fixed in version 6.0.0

PR #231, opened 2019-05-19
Input validation

gwenn:
Update #182


tailhook:
What is the state of this PR? Any help needed?

gwenn:
* `validate_while_typing` is not supported.
* I am not sure that the mutable line buffer will play well with the undo manager (to be checked).
* And if result is incomplete, issue #193 will most likely appear

tailhook:
I needed [these changes](https://github.com/tailhook/rustyline/commit/22e8a231dac97faaecd19f4be9419b8a56c7c11a) to make the current branch work, or else I get the errors on rust-1.40:
```
error[E0283]: type annotations needed: cannot resolve `std::string::String: std::convert::AsRef<_>`
   --> /work/target/.cargo/git/checkouts/rustyline-96f659ed5b4c338d/d15722d/src/lib.rs:627:43
    |                                                                          
627 |             editor.add_history_entry(line.as_ref());                 
    |                                           ^^^^^^
                                       
error[E0283]: type annotations needed: cannot resolve `std::string::String: std::convert::AsRef<_>`
   --> /work/target/.cargo/git/checkouts/rustyline-96f659ed5b4c338d/d15722d/src/history.rs:151:28
    |
151 |             self.add(line?.as_ref()); // TODO truncate to MAX_LINE
    |                            ^^^^^^                                        

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0283`.

```

Other than that at a glance, it works. I'm not sure how to check the undo manager you've outlined above. And yes #193 is there when you hit backspace on a new line.

What do we need to do to merge this other than fix the compilation bug? I think #193 is not new and can be handled afterward.

gwenn:
I am sorry.
I don't have much time to finalise this PR.
If we can implement [custom actions](#293) like [replxx](https://github.com/AmokHuginnsson/replxx/blob/master/src/replxx_impl.cxx#L1120), users can mutate the line buffers by invoking/combining several "standard" actions/commands like macros.
And be sure that both undo manager/kill ring are updated correctly.
Something like:
```rust
fn custom_action(ctx: &mut ActionContext) -> Result<ActionResult> {
    ctx.invoke(Cmd::Insert(1, ))?;
    ctx.invoke(Cmd::Move(Movement::EndOfLine))?;
    ...
    Ok(ActionResult::Continue)
}
```
Also we can refresh the screen only once (to avoid flickering).
But this implies refactoring the main loop and split many `State` methods.

Once we have an `ActionContext`, I guess we can do the same for validation: no direct mutation.
```rust
    fn validate(&self, ctx: &mut ValidationContext) -> Result<ValidationResult> {
        let _ = check(ctx.line());
        ctx.invoke(Cmd::Move(Movement::EndOfLine))?;
        ctx.invoke(Cmd::SelfInsert(1, ';'))?;
        ...
        Ok(...)
    }
```

tailhook:
Well, maybe I'm misunderstanding the issue here. What I want is just "continue input if the line doesn't contain semicolon". So I don't need a mutable buffer, or actions or whatever.

Maybe I would better just use loop along the lines of:
```rust
let initial = "";
loop {
    let line = readline_with_initial(">", (initial, ""));
    if is_full(line) { break line; }
    initial = line;
}
```

tailhook:
Returning from #311.

I propose we provision this:
```rust
enum ValidationResult {
  Valid,
  Incomplete,
  ModificationNeeded,
}
trait Validator {
  fn validate(self, data: &str) -> Result<ValidationResult>;
  fn correct(self, ctx:&mut ModificationContext);
}
```
But now implemement just this as MVP:
```rust
#[non_exaustive]
enum ValidationResult {
  Valid,
  Incomplete,
}
trait Validator {
  fn validate(self, data: &str) -> Result<ValidationResult>;
}
```

Looks quite easy to make this from the current PR.

Does it look sensible?

gwenn:
The original `ValidationResult` seems fine to me:
https://github.com/kkawakam/rustyline/pull/231/files#diff-e860543c88a7b150ab5ffca6b874971eR6
```
/// Input validation result
pub enum ValidationResult {
    /// Incomplete input
    Incomplete,
    /// Validation fails with an optional error message. User must fix the
    /// input.
    Invalid(Option<String>),
    /// Validation succeeds with an optional message (instead of https://github.com/kkawakam/rustyline/pull/169)
    Valid(Option<String>),
}
```
And if we want to provide auto-correction/auto-formatting/mutation in the future, it seems to me we need to introduce a `ValidationContext`:
```rust
    fn validate(&self, ctx: &mut ValidationContext) -> Result<ValidationResult> {
        let _ = check(ctx.line()); // immutable access
        // ctx.invoke(Cmd::Move(Movement::EndOfLine))?;
        // ctx.invoke(Cmd::SelfInsert(1, ';'))?; in the future, indirect mutation
        ...
        Ok(...)
    }
```

gwenn:
#193 has been fixed.

tailhook:
> And if we want to provide auto-correction/auto-formatting/mutation in the future

Yes, if you're okay with breaking changes. I've just proposed how to make it future proof. So we could introduce mutation later without breaking changes.

Anyway, I'm still not sure what needs to be done to merge this PR?

gwenn:
I am reluctant to provide an API (`fn validate(self, data: &str)` or `fn validate(&self, line: &mut LineBuffer)`) that we already know it will need to be broken/changed.

tailhook:
Fair enough. What are use-cases for modification in validator? (I don't see modification documented in [prompt-toolkit](https://python-prompt-toolkit.readthedocs.io/en/master/pages/asking_for_input.html#input-validation), but it's always unclear in python :smile:  )

By the way, a future-compatible alternative to [the coment above](https://github.com/kkawakam/rustyline/pull/231#issuecomment-570711200) would be something like this:
```rust
// Now
fn readonly_validate(&self, data: &str) -> Result<..>;
// Later
fn validate(&self, ctx: &mut ValidationContext) -> Result<..> {
  // default implementation
  self.readonly_validate(self, ctx.get_data()) 
}
```

Or should I implement #311 on top of [hack above](https://github.com/kkawakam/rustyline/pull/231#issuecomment-570570104) ?

gwenn:
```
// Now
fn readonly_validate(&self, data: &str) -> Result<..>;
// Later
fn validate(&self, ctx: &mut ValidationContext) -> Result<..> {
  // default implementation
  self.readonly_validate(self, ctx.get_data()) 
}
```
looks fine to me.

gwenn:
@tailhook Could you please review last commit ?
Thanks.

src/validate.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// Validation succeeds with an optional message
+    Valid(Option<String>),
+}
+
+pub struct ValidationContext<'i> {
+    i: &'i mut dyn Invoke,
+}
+
+impl<'i> ValidationContext<'i> {
+    pub(crate) fn new(i: &'i mut dyn Invoke) -> Self {
```
tailhook: Not sure why we need dyn trait here? It's not an a public constructor, so probably ValidationContext can do everything by itself.

src/edit.rs
```diff
@@ -217,6 +218,46 @@ impl<'out, 'prompt, H: Helper> State<'out, 'prompt, H> {
...
+                    if corrected || self.has_hint() || msg.is_some() {
+                        // Force a refresh without hints to leave the previous
+                        // line as the user typed it after a newline.
+                        self.refresh_line_with_msg(msg)?;
+                    }
+                    true
+                }
+                ValidationResult::Invalid(msg) => {
+                    if corrected || self.has_hint() || msg.is_some() {
+                        self.refresh_line_with_msg(msg)?;
```
tailhook: So the message here is highlighted the same as a hint, right? This should be documented in the Highlighter trait.

src/validate.rs
```diff
@@ -0,0 +1,81 @@
...
+use crate::Result;
+
+/// Input validation result
+#[non_exhaustive]
+pub enum ValidationResult {
+    /// Incomplete input
+    Incomplete,
+    /// Validation fails with an optional error message. User must fix the
+    /// input.
+    Invalid(Option<String>),
```
tailhook: I'm not sure about the validation message. It's fine in your example, but with bigger examples and copy-paste it might be ugly.

Like this (note, I also disallow empty backtick quotes):
```
> SELECT a, b, c, e as ``, f, g, h, i FROM test2 --< unexpected ``
```
Or like this (note, I've pasted the whole chunk of code, so validation is triggered at the end):
```
> SELECT
    a,
    b,
    COUNT(*) as ``
FROM
    test_table
;
 --< unexpected ``
```
Or this:
```
> ELECT
    a,
    b,
    COUNT(*) as ``
FROM
    test_table
; --< Expect: SELECT stmt
```

So to make that helpful we might need to yield position together with validation message and draw validation error in that position, perhaps like this:
```
> SELECT
    a,
    b,
    COUNT(*) as ``
                ^^ unexpected ``
FROM
    test_table
;
```

But I don't think it's worth doing now. The simpler implementation would probably return error span and highlight that span with a color while keeping the validation message at the end. What do you think?

gwenn: Yep. Currently, we cannot do much.
No [bottom bar](https://python-prompt-toolkit.readthedocs.io/en/master/pages/asking_for_input.html#adding-a-bottom-toolbar) !
No [external print](https://github.com/kkawakam/rustyline/pull/235) !
Should we merge ?
Thanks for your reviews.

tailhook: Yes. Let's merge and add more features later.

Issue #229, opened 2019-05-19
Externally triggered prompt redraw or write above the prompt


flxo:
Hi,

I really struggled with the title for this issue - so let me describe what I need:

I want to write a console application that provide s some kind of command line that is *always* present at the bottom of the terminal. The text above this is updated from events happening somewhere else and have no influence on the prompt.

```
abc
def
>> asdf
```
```
abc
def
ghi
>> asdf
```
Currently the `readline` blocks and I haven't found a way to kind of interrupt that and print some lines and resume the input operation. Is that possible with rustyline? 

I got this working with `linefeed` that allows to clone a handle to the prompter that implements write and internally synchronises writing to stdout. A example is [here](https://github.com/murarth/linefeed/blob/a6fbc376b242712084ce8efeb254fc096f25c1ce/examples/demo.rs#L83-L101). 

For sure this topics gets complicated when it comes to paged display of the history etc...
Did this topic already raise in the past and if yes - what's the conclusion?

Thanks,

flxo

gwenn:
I don't know how to do this!
I should investigate how [`python-prompt-toolkit`](https://python-prompt-toolkit.readthedocs.io/en/stable/pages/asking_for_input.html#prompt-in-an-asyncio-application) or `linefeed` do this.
I will look at your example.

gwenn:
See similar topics:
 * #48
 * #199
 * #126

Async/non-blocking input but sync/coherent output.

neonphog:
+1 I need this too. Looks like https://github.com/Dhole/rustylinez is trying to add this functionality, but didn't compile for me

gwenn:
Just for reference:
- haskeline [getExternalPrint](http://hackage.haskell.org/package/haskeline-0.7.5.0/docs/System-Console-Haskeline.html#g:6)
- python prompt toolkit [patch_stdout](https://python-prompt-toolkit.readthedocs.io/en/master/pages/asking_for_input.html?highlight=patch_stdout#prompt-in-an-asyncio-application) ([example](
https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/examples/prompts/patch-stdout.py))
- replxx [print](https://github.com/AmokHuginnsson/replxx/blob/master/include/replxx.hxx#L223) ([example](https://github.com/AmokHuginnsson/replxx/blob/master/examples/cxx-api.cxx#L249))
[...]

PR #235, opened 2019-06-02
External print

gwenn:
See #229

```sh
cargo run --example external_print
```

To do:
* ~~Windows platform (only one printer supported, and when the printer is dropped we still wait)~~
* Check interaction with history search, completion, ...


gwenn:
https://github.com/ghc/ghc/blob/master/libraries/base/cbits/inputReady.c#L325
> nightmare.  A Console Handle will appear to be ready
> ...

gwenn:
Something is broken with last merge.
Sorry.

tailhook:
It works fine, except for the following issue:

If I don't put a newline to the output or the external printer like `printer.write(" ")`, rustyline hangs.

While I understand that partial line would interfere with input prompt, it's better to have clobbered prompt (which will eventually fix on next keystrokes), than hanging.

flxo:
Hi.

I tried this branch and discovered the following: When using the printer from different threads, the blocking `readline` call somehow interferes with the `std::io::Write` implementation produced by `create_external_printer`. 

Here's a (hacky) minimal example with two thread sending/listening on channels. One thread performs the `readline` and the other one just prints.

```rust
use std::{io::Write, sync::mpsc, thread};
use rustyline::Editor;

fn main() {
    let mut rl = Editor::<()>::new();

    let mut printer = rl.create_external_printer().unwrap();
    let print = {
        let (tx, rx) = mpsc::channel::<String>();
        thread::spawn(move || {
            while let Ok(line) = rx.recv() {
                printer.write_all(line.as_bytes()).expect("Failed to write");
                printer.write_all(b"\n").expect("Failed to write");
                printer.flush().expect("Failed to flush");
            }
        });
        tx
    };

    let readline = {
        let (tx, rx) = mpsc::channel::<String>();
        thread::spawn(move || {
            loop {
                match rl.readline(">> ") {
                    Ok(line) => {
                        rl.add_history_entry(line.as_str());
                        tx.send(line).ok();
                        //std::thread::sleep(std::time::Duration::from_secs(2));
                    },
                    _ => break,
                }
            }
        });
        rx
    };

    loop {
        match readline.recv() {
            Ok(line) => {
                for n in 0..10 {
                    print.send(format!("{} {}", n, line)).ok();
                }
            }
            Err(_)=> break,
        }
    }
}
```

With `rustyline = { git = "https://github.com/gwenn/rustyline.git", branch = "external-print", features = ["with-fuzzy"] }`

The output is e.g:

```
>> AAA
0 AAA
>> BBB
1 AAA
2 AAA
3 AAA
4 AAA
5 AAA
6 AAA
7 AAA
8 AAA
9 AAA
0 BBB
>> CCC
1 BBB
2 BBB
3 BBB
4 BBB
5 BBB
6 BBB
7 BBB
8 BBB
9 BBB
0 CCC
>>
```

The 9 `AAA`s are printed after `BBB` is entered and return is hit. Looks like some buffer within the printer is not flushed correctly or the writing is blocked.

Originally I'm using `rustyline` in a Tokio environment and run the `readline` and the printing in `spawn_blocking` Tokio tasks. The behaviour is the same as with the std threads. 

For testing, I added a stupid `sleep` for 5 secons in the readline thread after each input and the output is printed as expected.

gwenn:
> I tried this branch and discovered the following: When using the printer from different threads, the blocking `readline` call somehow interferes with the `std::io::Write` implementation produced by `create_external_printer`.

On unix or windows ?

flxo:
> > I tried this branch and discovered the following: When using the printer from different threads, the blocking `readline` call somehow interferes with the `std::io::Write` implementation produced by `create_external_printer`.
> 
> On unix or windows ?

Sorry forgot to add this. I tested on a x86 Linux and Mac.

gwenn:
I will try to have a look this week-end.

gwenn:
@flxo Ok, we should not use a `BufReader` [here](https://github.com/kkawakam/rustyline/pull/235/files#diff-bc8fd5ba686c57cce3f8eefc70d2b5419468e9c84b5d1923095584f8dabb010bR1222). Because buffered lines are not "selectable" anymore...

gwenn:
@flxo Your issue should have been fixed by 0a5f0a4 on unix platform.

flxo:
@gwenn Thanks for the update. I tried your patches and the issue is gone. 

In my software I have a (hacky) `std::io::Write` impl around the `rustyline::ExternalPrinter` which allows me to use `std::io::stdout` and the rustyline interface dependant on the use case (iteractive or not).
Would it be possible to provide `std::io::Write` directly?

gwenn:
> @gwenn Thanks for the update. I tried your patches and the issue is gone.
> 
> In my software I have a (hacky) `std::io::Write` impl around the `rustyline::ExternalPrinter` which allows me to use `std::io::stdout` and the rustyline interface dependant on the use case (iteractive or not).
> Would it be possible to provide `std::io::Write` directly?

You should be able to implement an `ExternalPrinter` that writes to `stdout`, no ?

flxo:
@gwenn My idea is to do something like

```rust
impl<T> std::io::Write for T where T: ExternalPrinter {
   ...
}
```

The prints are some via the `writeln!` family and with the `impl` of `std::io::Write` I can decided whether to use `std::io::Stdout` for non interactive usage (there's no rustyline::Editor around) and the thing returned from `create_external_printer`

```rust
let out = if interactive {
  Box::new(rl.create_external_printer()?) as Box<dyn std::io::Write>
} else {
  Box::new(std::io::stdout()) as Box<dyn std::io::Write>
}

writeln!(out, "test");
```

gwenn:
The `ExternalPrinter` is now message oriented (`\n` is not meaningful anymore).
And `Write` is bytes/stream oriented.
So I would prefer that you implements `Write` on your own ExternalPrinter wrapper.

bartlomieju:
Is there any chance this PR could be landed? We use `rustyline` for REPL in Deno and this feature would greatly improve the experience for streaming output (https://github.com/denoland/deno/issues/8049), currently the output is interleaved with prompt line.

gwenn:
> Is there any chance this PR could be landed?

Current implementation is incomplete on windows.
> EDIT: Actually not a huge priority, we're gonna experiment with different crate for this purpose https://docs.rs/reedline

#579

vi:
> Current implementation is incomplete on windows.

Can UNIX-only implementation land first then?

gwenn:
Could you please give it another try (windows implementation should be fine now) ?
Thanks.

PR #266, opened 2019-08-15
Add an example to read password

gwenn:


madprops:
Thanks for this. Seems I'm having some problem when I make an editor instance that uses a config:

```
    let h = MaskingHighlighter {masking: false};
    let mut editor = Editor::<()>::with_config(config);
    editor.set_helper(Some(h));
```

It complains on Some(h): **expected (), found struct `structs::MaskingHighlighter`**

madprops:
Oh my bad I was init the editor wrong. **Editor::with_config(config);** worked.

madprops:
Works great, thanks!

Issue #278, opened 2019-08-20
Circular dependencies


gwenn:
`rustyline-derive` depends on `rustyline` traits.
`rustyline` examples depends on `rustyline-derive`.

If some new trait is introduced (like [here](https://github.com/kkawakam/rustyline/pull/231/files#diff-b4aea3e418ccdb71239b96952d9cddb6R651)), I don't see how we will publish !

Move examples in `rustyline-derive` ?

PR #279, opened 2019-08-20
Fix circular dependencies

gwenn:
https://stackoverflow.com/questions/54088182/cyclic-package-dependency-while-implementing-proc-macro
> Your derive crate should not depend of anything because it only
handles code generation.


gwenn:
Fix #278

PR #282, opened 2019-08-27
Support bell style

### LINKED ISSUE #281: Option to disable bell sound
The bell rings too often in Windows, when using tab completion

I am using nushell, and filed a ticket on them to disable the bell, but apparently rustyline is the culprit https://github.com/nushell/nushell/issues/446 

Rustyline could support bell-style as seen elsewhere, but option to kill that altogether from using application would be nice.

gwenn:
Should fix #281

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -4,7 +4,7 @@ API
 Async (#126)
 
 Bell
-- [ ] bell-style
+- [X] bell-style
 
 Color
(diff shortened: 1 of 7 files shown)


vivainio:
Thanks for the prompt implementation. I'll try upgrading the upgraded version in nushell after you cut the next release

PR #293, opened 2019-10-20
Conditional Bind Sequences

### LINKED ISSUE #269: Conditional Escape
I use `editor.bind_sequence(KeyPress::Esc, Cmd::Interrupt);`
Which works. I'm wondering if it could be possible to make it clear text if there's some written text, and if not Interrupt..

gwenn:
Make possible to return different command depending on input state.

Try to fix #269.
No breaking change but a `Cmd` enum with `Custom(Action)` entry which can return a `Cmd` looks bad...


src/keymap.rs
```diff
@@ -13,6 +14,58 @@ use crate::tty::{RawReader, Term, Terminal};
...
 pub type RepeatCount = usize;
 
+/// Custom dynamic action
+#[derive(Clone)]
+pub struct Action {
+    pub name: String,
+    /// Takes the currently edited `line` with the cursor `pos`ition and
+    /// returns the command to be performed or `None` to perform the default
+    /// one.
+    pub action: fn(line: &str, pos: usize, ctx: &ActionContext) -> Option<Cmd>,
```
cchalmers: I feel like the `line` and `pos` should be part of the context.

An `fn` is pretty restrictive because it doesn't allow adding any of your own context. Having something like
```
Arc<Box<dyn Fn(&ActionContext) -> Option<Cmd>>>
```
would be much more flexible.

Also I think it would be better to not make this public to allow non-breaking changes in the future.

src/keymap.rs
```diff
@@ -13,6 +14,58 @@ use crate::tty::{RawReader, Term, Terminal};
...
+    }
+}
+
+impl PartialEq for Action {
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name
+    }
+}
+
+pub struct ActionContext<'r> {
```
cchalmers: You've made the struct public but it's isn't exposed anywhere so no documentation gets generated for it.

src/lib.rs
```diff
@@ -53,7 +53,9 @@ use crate::edit::State;
 use crate::highlight::Highlighter;
 use crate::hint::Hinter;
 use crate::history::{Direction, History};
-pub use crate::keymap::{Anchor, At, CharSearch, Cmd, Movement, RepeatCount, Word};
+pub use crate::keymap::{
+    Action, Anchor, At, CharSearch, Cmd, InputMode, Movement, RepeatCount, Word,
```
cchalmers:
```suggestion
    Action, ActionContext, Anchor, At, CharSearch, Cmd, InputMode, Movement, RepeatCount, Word,
```

src/keymap.rs
```diff
@@ -13,6 +14,58 @@ use crate::tty::{RawReader, Term, Terminal};
 /// The number of times one command should be repeated.
 pub type RepeatCount = usize;
 
+/// Custom dynamic action
+#[derive(Clone)]
+pub struct Action {
+    pub name: String,
```
cchalmers: Can you say what this name is used for? It's not clear to me.

src/keymap.rs
gwenn: Most of the time, you just need to check `line` and/or `pos`. So if you can access them directly, it seems better/easier.
And could you provide a case where an action is not stateless ?

src/keymap.rs
gwenn: My bad.

src/lib.rs
gwenn: Indeed.

src/keymap.rs
gwenn: It is used only for tracing / debugging.

gwenn:
@cchalmers We still need to support composite command. For example, clear whole line and interrupt.
Either by mutating the line buffer directly like [linefeed](https://github.com/murarth/linefeed/blob/master/examples/enter_function.rs#L41).
Or by invoking multiple commands like [replxx](https://github.com/AmokHuginnsson/replxx/blob/master/examples/cxx-api.cxx#L142).

And last, we need to fix/move [this](https://github.com/kkawakam/rustyline/blob/master/src/edit.rs#L90)

src/keymap.rs
cchalmers: For example I've been playing around with having a key bindings to insert bits of text with [`skim`](https://github.com/lotabout/skim). The bits of text can be dependent on the rest of the program.

gwenn:
@cchalmers I mean that if your custom action replaces some text, we cannot just do a pattern matching anymore (Cmd::Custom doesn't match Cmd::Replace)...

src/keymap.rs
```diff
@@ -394,16 +468,13 @@ impl InputState {
...
-                return Ok(if cmd.is_repeatable() {
-                    cmd.redo(Some(n), wrt)
-                } else {
-                    cmd.clone()
-                });
-            }
+
+        if let Some(cmd) = self.custom_binding(wrt, key) {
+            return Ok(if cmd.is_repeatable() {
+                cmd.redo(Some(n), wrt)
```
tailhook: This replaces number returned from the custom action. I.e. I can't return `Move(BackwardChar(10)`.

src/keymap.rs
tailhook: I don't understand why `Action::action` contains a function. Rather than `Action` being a trait.

src/keymap.rs
gwenn: Right. This should not be done for `Action`, only for simple bindings.

src/keymap.rs
gwenn: Mainly because `Action` is `Sized`.

tailhook: We have just merged `Validation` where anyone would use `line` for sure, but it only has `ValidationContext` in the signature. Well, the reason for validator is probably because line is mutable there. But I think having similar signature is nice anyway.

tailhook: `Box<dyn Action>` is also sized. And I don't think it can be a performance issue here.

nu8:
https://github.com/kkawakam/rustyline/issues/306

89z:
https://github.com/kkawakam/rustyline/issues/306

89z:
https://github.com/kkawakam/rustyline/issues/306

Issue #294, opened 2019-10-22
Crash when searching for a value


vsajip:
This code:
```
fn main() {
    use std::io::{stdin, stdout, Write};

    use rustyline::error::ReadlineError;
    use rustyline::Editor;

    loop {
        let mut editor = Editor::<()>::new();
        if editor.load_history("history.log").is_err() {
        }
        let result = editor.readline("Enter expression (ctrl-C to quit): ");
        match result {
            Ok(mut line) => {
                editor.add_history_entry(line.as_str());
                line = line.trim().to_string();
                if line.len() > 0 {
                    println!("You entered: {}", line);
                }
            },
            Err(ReadlineError::Interrupted) => {
                break
            },
            Err(ReadlineError::Eof) => {
            },
            Err(err) => {
                println!("Error: {:?}", err);
                break
            }
        }
        editor.save_history("history.log").unwrap();
    }
}
```

Fails when run if you do a `ctrl-R` to search for a previous value (which _is_ found). The failure enters when you press Enter on the found value. Console listing with backtrace (Ubuntu-x64, rust 1.38.0):

```
$ RUST_BACKTRACE=full target/debug/expr_repl
Enter expression (ctrl-C to quit): abc
You entered: abc
(reverse-i-search)`ab': abc           thread 'main' panicked at 'assertion failed: self.layout.cursor <= self.layout.end', /home/vinay/.cargo/registry/src/github.com-1ecc6299db9ec823/rustyline-5.0.3/src/edit.rs:125:13
stack backtrace:
   0:     0x5565d5ee898b - backtrace::backtrace::libunwind::trace::h89fcc71e59e3bc5b
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.34/src/backtrace/libunwind.rs:88
   1:     0x5565d5ee898b - backtrace::backtrace::trace_unsynchronized::h0bad9be1379e729a
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.34/src/backtrace/mod.rs:66
   2:     0x5565d5ee898b - std::sys_common::backtrace::_print::hd3382a1f33c473da
                               at src/libstd/sys_common/backtrace.rs:47
   3:     0x5565d5ee898b - std::sys_common::backtrace::print::h0ec6f03cfb8e76a6
                               at src/libstd/sys_common/backtrace.rs:36
[...]

PR #295, opened 2019-10-22
Fix for #294 - reverse search crash - the tests still pass, but library doesn't crash

vsajip:
I commented out the assertion which was causing the problem reported in #294, and added the simple program which was causing the crash.


gwenn:
The assert should not be removed.
Would you mind testing PR #296 ?
And close this PR.
Thanks.

PR #298, opened 2019-10-30
Add a convenience SimpleEditor type alias

RadicalZephyr:
I think it's common enough that people use `Editor<()>` that providing an easy to remember type alias for it makes sense.


gwenn:
See https://docs.rs/rustyline/latest/rustyline/type.DefaultEditor.html

PR #324, opened 2020-01-24
Add LineUp/LineDown movement commands

### LINKED ISSUE #317: Cursor up in multi-line mode doesn't move to the previous line
In most other shells I've used cursor keys work up and down when inside the multi-line string and only switch history on the first and last lines of that input respectively. Is there a configuration that allows that in rustyline?

tailhook:
This PR also changes `j/k` and `+/-` in vi mode to use these movements

For the vi command mode, Ctrl+N and Ctrl+P switch history. And Up/Down array keys switch history in insert mode.

The emacs mode is left intact in this PR.

This partially fixes #317

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -508,8 +508,6 @@ fn readline_edit<H: Helper>(
                 let c = rdr.next_char()?;
                 s.edit_insert(c, 1)?
             }
-            #[cfg(not(unix))]
-            Cmd::QuotedInsert => {}
             Cmd::Yank(n, anchor) => {
                 // retrieve (yank) last item killed
                 let mut kill_ring = editor.kill_ring.lock().unwrap();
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -597,7 +605,14 @@ fn readline_edit<H: Helper>(
...
             }
-            Cmd::Noop | _ => {
+            #[cfg(not(unix))]
+            Cmd::Suspend => {}
+            Cmd::Move(Movement::WholeLine) => {} // not an actual movement
+            Cmd::Complete | Cmd::CompleteBackward | Cmd::CompleteHint |
+            Cmd::ForwardSearchHistory | Cmd::ReverseSearchHistory |
+            Cmd::Abort | Cmd::Insert(_, _) | Cmd::SelfInsert(_, _) |
+            Cmd::Noop | Cmd::Unknown
+            => {
```
gwenn: Is this change related to https://rust-lang.github.io/rust-clippy/master/index.html#wildcard_in_or_patterns ?
Why not just replace `Cmd::Noop | _` by `_` ?

tailhook: This way it's much easier to add a new command. The compiler shows all the places where a new command should be added, including this match statement. And no it's not to satisfy Clippy.

I can revert this change if you don't think it makes sense.

gwenn: If you don't mind, I would prefer that you revert this change.
Thanks.

tailhook: Reverted. Sorry for long delay. (I've reverted in a different commit, I expect you to squash changes when merging. Let me know if you want me to squash changes.

Issue #323, opened 2020-01-24
Feature Request: Fuzzy CompletionType option


Jacobious52:
It would be neat and useful to have another `CompletionType` option that uses fuzzy searching, similar to functionality that provides [fzf](https://github.com/junegunn/fzf). I have a [branch](https://github.com/Jacobious52/rustyline/blob/fuzzy-completion/src/lib.rs#L189) that uses [skim](https://github.com/lotabout/skim) to filter and select the completion as a proof of concept, but want to make sure this is a feature that would fit in rustyline. I imagine it would be an optional cargo feature as it adds more dependencies.

Here's an example using [nushell](https://github.com/nushell/nushell) which uses rustyline with my branch with fuzzy completion.
![image](https://user-images.githubusercontent.com/12645162/73043962-a538f000-3ebb-11ea-89e5-a5bafdf1e2ca.png)

There's still some polish and some undesired behaviour, for example my current implementation adds a new line to the prompt. Not sure how to make it replace the existing line instead.

If this is a feature that would be useful, I'm happy to PR with some guidance on polishing the implementation.

This fuzzy search would also be useful for `(reverse-i-search)` mode, but I haven't gotten that far in the implementation yet.

Cheers

Jacobious52:
Closed with #326

PR #326, opened 2020-01-25
Add skim as optional fuzzy completion option for unix

Jacobious52:
Implements #323 

This PR adds optional support for Fuzzy Completion mode using [skim](https://github.com/lotabout/skim) as the backend. Unfortunately skim uses [tuikit](https://github.com/lotabout/tuikit) for it's terminal UI which doesn't support Windows systems yet, so I've included compile conditionals for `unix` as well as the optional feature `with-fuzzy` for the time being.

**To demo:**

- Use CompletionType::Fuzzy, instead of List or Circular. e.g. replace the [example here](https://github.com/kkawakam/rustyline/blob/master/examples/example.rs#L71)
- Run: `cargo run --example example --features=with-fuzzy`
- Use tab as normal to autocomplete the file system.

**Snapshots:**

Simple autocomplete while filtering:

image

Complete subdirectory and add selected completion

image
image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ memchr = "2.0"
 [target.'cfg(unix)'.dependencies]
 nix = "0.16"
 utf8parse = "0.2"
+skim = { version = "0.7", optional = true }
 
 [target.'cfg(windows)'.dependencies]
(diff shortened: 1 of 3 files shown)


gwenn:
Thanks.

PR #329, opened 2020-02-04
Add smart up and down commands

tailhook:
While not ubiquitous, it's quite common and intuitive for `Up/Down` arrow keys to navigate between lines in the multi-line prompt, and only switch history if the cursor is on the first/last (or the only) line of the input.

This PR adds both the commands that do that and change Up/Down arrow keys. Note: Ctrl+N/P remain the same as well as `j/k` in vim mode.


gwenn:
Thanks.

PR #330, opened 2020-02-04
Add AcceptOrNewLine command

tailhook:
This command might be used to insert a newline by default if cursor is
not on the end of line.

This is useful for where multiple lines are not occassional but are used
most of the time (i.e. SQL, Lisp...). The `AcceptLine` only inserts
newline when the input is non-valid, so inserting a thing into the fully
valid chunk of text is cumbersome.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -281,6 +281,7 @@ fn page_completions<C: Candidate, H: Helper>(
                 && cmd != Cmd::SelfInsert(1, ' ')
                 && cmd != Cmd::Kill(Movement::BackwardChar(1))
                 && cmd != Cmd::AcceptLine
+                && cmd != Cmd::AcceptOrInsertLine
             {
                 cmd = s.next_cmd(input_state, rdr, false)?;
             }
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -586,6 +587,26 @@ fn readline_edit<H: Helper>(
...
+                {
+                    editor.term.cursor = s.layout.cursor.col;
+                }
+                if s.has_hint() || !s.is_default_prompt() {
+                    // Force a refresh without hints to leave the previous
+                    // line as the user typed it after a newline.
+                    s.refresh_line_with_msg(None)?;
+                }
+                // Only accept value if cursor is at the end of the buffer
+                if s.validate()? && s.line.is_end_of_input() {
```
gwenn: I am not sure but we don't need to duplicate code:
```rust
Cmd::AcceptLine | Cmd::AcceptOrInsertLine => {
                #[cfg(test)]
                {
                    editor.term.cursor = s.layout.cursor.col;
                }
                // Accept the line regardless of where the cursor is.
                if s.has_hint() || !s.is_default_prompt() {
                    // Force a refresh without hints to leave the previous
                    // line as the user typed it after a newline.
                    s.refresh_line_with_msg(None)?;
                }
                 s.edit_move_end()?;
                 if s.validate()? && (cmd == Cmd::AcceptLine || s.line.is_end_of_input()) {
                     break;
                 } else {
                     s.edit_insert('\n', 1)?;
                 }
                 continue;
}
```

tailhook: Okay

tailhook:
Fixed code duplication and amended the commit.

gwenn:
Thanks.

PR #333, opened 2020-02-12
Split KeyPress into separate Key and Modifier types.

### LINKED ISSUE #318: Unable to do Meta-Left or Meta-Right
The current implementation of [`KeyPress`](https://docs.rs/rustyline/6.0.0/src/rustyline/keys.rs.html#26) only allows Meta to take a `char` rather than another `KeyPress`, which also has a `KeyPress::Char(char)` variant, and as such it disallows multi char control sequences. On OSX if one wishes to implement Ctrl+Left or Ctrl+Right functionality for a terminal shell the equivalent is Meta+Left and Meta+Right as the prior functionality is bound to switching workspaces.

Would having it take a `KeyPress` be reasonable? I see that things like `KeyPress::ControlLeft` exist but this seems less flexible rather than allowing some command modifier taking another `KeyPress` or multiple of them. Maybe I'm missing some architectural choice as to why it is this way, but I couldn't find anything in the issues or git log that would help me figure out why and have probably overlooked something here.

thomcc:
Also refactor the escape seq. parser to be data-driven, making it easy to add support for these.

Fixes #318 , maybe some others. Big breaking change, but I tried to make the changes not as painful as they naively would be. 

I need to test more, it's late, no rush on this review (It's big, but most of the parts were for updating code with the changes), it just bugged me is all.

### DIFF
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -327,7 +327,7 @@ pub fn char_to_key_press(c: char) -> KeyPress {
         '\x07' => KeyPress::ctrl('G'),
         '\x08' => KeyPress::BACKSPACE, // '\b'
         '\x09' => KeyPress::TAB,       // '\t'
-        '\x0a' => KeyPress::ctrl('J'), // '\n' (1)
+        '\x0a' => KeyPress::ctrl('J'), // '\n' (10)
         '\x0b' => KeyPress::ctrl('K'),
         '\x0c' => KeyPress::ctrl('L'),
         '\x0d' => KeyPress::ENTER, // '\r' (13)
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -1,19 +1,14 @@
 //! Key constants
 
-// #[non_exhaustive]
+/// A key, independent of any modifiers. See KeyPress for the equivalent with modifers.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub enum KeyPress {
+pub enum Key {
+    Char(char),
     UnknownEscSeq,
     Backspace, // Ctrl('H')
     BackTab,
     BracketedPasteStart,
     BracketedPasteEnd,
-    Char(char),
-    ControlDown,
-    ControlLeft,
-    ControlRight,
-    ControlUp,
-    Ctrl(char),
     Delete,
     Down,
     End,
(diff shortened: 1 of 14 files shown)


thomcc:
I've tested this on various terminals and it seems to work well.

src/tty/unix.rs
```diff
@@ -126,13 +127,324 @@ impl Read for StdinRaw {
...
+    // b"\x1b[18;2~" as a binding, `prefixes` will store `\x1b`, `\x1b[`,
+    // `\x1b[1`, `\x1b[18`, `\x1b[18;`, and so on. We also use it to detect if
+    // there's an ambiguous binding during startup.
+    //
+    // In an ideal world we'd use some prefix tree structure, but I surprisingly
+    // couldn't find one that actually supported an `item_with_prefix_exists`
+    // (or similar) function.
+    //
+    // This exists so we know when to emit an "unknown escape code" complaint.
+    prefixes: HashSet<EscapeSeq>,
```
gwenn: https://docs.rs/radix_trie/0.1.6/radix_trie/struct.Trie.html#method.subtrie
https://docs.rs/qp-trie/0.7.4/qp_trie/struct.Trie.html#method.subtrie_str
?

src/keys.rs
```diff
@@ -23,68 +18,451 @@ pub enum KeyPress {
...
+        '\x01' => KeyPress::ctrl('A'),
+        '\x02' => KeyPress::ctrl('B'),
+        '\x03' => KeyPress::ctrl('C'),
+        '\x04' => KeyPress::ctrl('D'),
+        '\x05' => KeyPress::ctrl('E'),
+        '\x06' => KeyPress::ctrl('F'),
+        '\x07' => KeyPress::ctrl('G'),
+        '\x08' => KeyPress::BACKSPACE, // '\b'
+        '\x09' => KeyPress::TAB,       // '\t'
+        '\x0a' => KeyPress::ctrl('J'), // '\n' (1)
```
gwenn: (10)

gwenn:
Thanks for all you do.
I just need some time to review all these changes...

thomcc:
No worries. The majority of it was done via find and replace FWIW, since it was basically a big rename of a lot of stuff.

src/tty/unix.rs
```diff
@@ -126,13 +128,305 @@ impl Read for StdinRaw {
...
+
+    // Ideally some of this would be read out of terminfo, but... that's a pain
+    // and terminfo doesn't have entries for everything, and sometime's it's
+    // just wrong anyway (*cough* iTerm *cough*). So instead we just add
+    // more-or-less everything we care about to the set of bindings. That said,
+    // this is actually fine. These are the strings that are coming out of the
+    // terminal, so long as two terminals don't use the same strings to mean
+    // different things, it won't be a problem that we have so many items in our list.
+    fn add_common_unix(&mut self) {
+        // Ansi, vt220+, xterm, ...
```
gwenn: To be checked with https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/prompt_toolkit/input/ansi_escape_sequences.py#L68-L312

thomcc: I hadn't seen that, thanks.

Taking a quick look, it seems like they have a few things we're missing, but for the stuff we do support I think we are doing better than them (they don't seem to handle some of the ways iTerm / Terminal.app are configured, as well as some of the xterm sequences we have).

- Our code doesn't handle function keys beyond F12. They go to F24, in theory terminals support up to F64 but... I've never seen a keyboard with more than 12, and the old code stopped at F12 I think.
- They support some cool stuff like scrolling that we should add.
- They support some `mintty`-specific escape sequences, which I wasn't aware of, and would be probably easy to add.

These seem like good things to add, but none of them (AFAIK?) were supported prior to my changes either, so at least they aren't regressions.

That said, I think we should definitely ensure we aren't missing anything, and that page looks good for that. I'm glad to see they have a similar approach of "just listen for everything" instead of following terminfo, which was wrong in many cases I found.

If you want to check my work further, I put this set together using a combination of:

1. Installing a fully updated `terminfo` for various terminals people seem to use and ensuring we support the sequences those advertise in their terminfo (both according to asking around and playing with the https://pkgstats.archlinux.de/ api some).
    - Often the actual terminfo for their default `TERM` variable is lacking, but a combination of seeing if the [terminfo source](https://raw.githubusercontent.com/mirror/ncurses/47d2fb4537d9ad5bb14f4810561a327930ca4280/misc/terminfo.src)
    - That said, this turned out to be wrong or incomplete in a lot of cases, so I went with just listening for all the escapes at once.

2. Reading the XTerm control sequence documentation: https://invisible-island.net/xterm/ctlseqs/ctlseqs.html. This helped a lot, since xterm is at this point practically the standard.

3. A lot of manual testing by installing various terminals, firing up `fish_key_reader -c` (Extremely helpful for this, even if you don't use fish -- prints out exactly what bytes each sequence sends), and seeing what sequences are written in response to different actions.

4. And then leaving in any sequence that the old code handled, even if I couldn't find a source for it, so long as it would be harmless (which they all seemed to be.

src/keys.rs
```diff
@@ -23,68 +18,451 @@ pub enum KeyPress {
...
+    }
+
+    #[inline]
+    pub const fn shift(self) -> KeyPress {
+        self.with_mods(KeyMods::SHIFT)
+    }
+
+    #[inline]
+    pub const fn ctrl(self) -> KeyPress {
+        self.with_mods(KeyMods::CTRL)
```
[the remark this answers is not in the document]
thomcc: Hrmm... We can't prevent this without encapsulating stuff more (which would break our ability to `match` on a key press.

The old code had the same problem though -- Nothing stopped a user for looking at `KeyPress::Ctrl('M')`. Events coming out of `rustyline` should always get this right though, but I guess it could confuse people adding bindings.

The only thing I can think is to add some comments warning about it, I think in practice it's pretty unlikely to be hit tho, unless I'm missing something?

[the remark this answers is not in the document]
thomcc: Actually, I found out that some terminals can provide sequences distinguishing these via CSI u sequences: http://www.leonerd.org.uk/hacks/fixterms/.

So I think it's better as-is? I should make sure these are supported here though.

gwenn:
Conflicts resolved in https://github.com/thomcc/rustyline/pull/1

src/tty/unix.rs
bjorn3: > Our code doesn't handle function keys beyond F12. They go to F24, in theory terminals support up to F64 but... I've never seen a keyboard with more than 12, and the old code stopped at F12 I think.

You can use `tmux send-keys` to send the higher function keys if you are using tmux.

src/tty/unix.rs
```diff
@@ -147,228 +441,47 @@ impl PosixRawReader {
...
-                        ('2', 'A') => KeyPress::ShiftUp,
-                        ('2', 'B') => KeyPress::ShiftDown,
-                        ('2', 'C') => KeyPress::ShiftRight,
-                        ('2', 'D') => KeyPress::ShiftLeft,
-                        _ => {
-                            debug!(target: "rustyline",
-                                   "unsupported esc sequence: ESC [ 1 ; {} {:?}", seq4, seq5);
-                            KeyPress::UnknownEscSeq
+                EscapeSearchResult::IsPrefix => {}
+                EscapeSearchResult::NoMatch => {
```
gwenn, line 465: I am afraid we need to read the complete escape sequence before returning `UnknownEscSeq`.
Otherwise, we will print its tail.
I guess, we should better use something like `vte` crate.

src/tty/windows.rs
```diff
@@ -228,14 +198,16 @@ impl RawReader for ConsoleRawReader {
...
-                        key = KeyPress::BackTab;
-                    } else if key == KeyPress::Char(' ') && ctrl {
-                        key = KeyPress::Ctrl(' ');
+                    if key == KeyPress::TAB && shift {
+                        key = KeyPress::BACK_TAB;
+                    } else if key == KeyPress::from(' ') && ctrl {
+                        key = KeyPress::ctrl(' ');
                     }
+                    // XXX should this be key.with_mods(mods)? Leaving it as-is
+                    // for now because it seems deliberate.
```
gwenn, line 210: We should do something like https://github.com/crossterm-rs/crossterm/blob/master/src/event/sys/windows/parse.rs#L120.
To support something like Ctrl-Alt-A.

thomcc:
Sorry for vanishing here, and thanks for taking over on the PR. I think there are a few sequences still missing from this (aside from ESC ESC, which is added by #517), but at some point I'll try to look into it, and if so, file a bug about them concretely.

PR #334, opened 2020-02-12
Bugfixes/consistency improvements for multiline stuff

thomcc:
- Movement::{BeginningOfLine,EndOfLine,etc} now actually go to beginning/end of the line.
- Movement::BeginningOfBuffer, etc was added to get the original functionality back.
- Fix some off-by-one bugs in the line movement code.
- Make it easier to test the multiline functionality by adding a matching bracket validator, and using it in the examples.

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -5,18 +5,21 @@ use rustyline::config::OutputStreamType;
 use rustyline::error::ReadlineError;
 use rustyline::highlight::{Highlighter, MatchingBracketHighlighter};
 use rustyline::highlight::{PromptInfo};
+use rustyline::validate::{self, Validator, MatchingBracketValidator};
 use rustyline::hint::{Hinter, HistoryHinter};
 use rustyline::{Cmd, CompletionType, Config, Context, EditMode, Editor, KeyPress};
-use rustyline_derive::{Helper, Validator};
+use rustyline_derive::Helper;
 
-#[derive(Helper, Validator)]
+#[derive(Helper)]
 struct MyHelper {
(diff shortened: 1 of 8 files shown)


gwenn:
zsh seems to escape only '\n':
```
echo "\test\
t"
echo "\test\\
t"
```

gwenn:
But zsh is buggy if line ends with a backslash:
```sh
$ echo test\\
$ # a space is appended in history 
```

thomcc:
That part is just so that if you have a multiline history item, after saving/loading it doesn't come back as individual lines.

gwenn:
With bash, Ctrl-A/Ctrl-E|Ctrl-U/Ctrl-K moves|kills to begin/end of buffer even with multiline input.
Maybe we should keep the current behaviour ?

thomcc:
It's a mix with what does what -- fish-shell and... every text editor supporting these have it go to the beginning of the line.

Also, as an anecdote: I use these very commonly and find going to the beginning of the buffer very frustrating (as it behaves differently than text editors).

gwenn:
Ok.

tailhook:
Anything left for this PR to get merged (except resolving conflicts of course)? We already use the code from the branch, and are pretty happy with it.

thomcc:
Sorry, I was mistaken that I'd have time last weekend, but I have nothing else planned for this weekend/tomorrow afternoon and will rebase it.

thomcc:
This should be good now.

gwenn:
Thanks.

PR #336, opened 2020-02-21
Fix incorrect cfg-if dependency declaration.

vsajip:
There is an incorrect dependency declaration for `cfg-if` in `Cargo.toml`. See discussion here:

https://users.rust-lang.org/t/library-compiles-and-runs-tests-but-fails-compilation-when-used-as-a-dependency/38463/2

This PR just adjusts the `cfg-if` dependency to be `0.1.6` rather than `0.1`.


gwenn:
Thanks.

PR #338, opened 2020-02-25
Always use rectangular prompt for multi-line input

### LINKED ISSUE #311: Multi-line prompts
I want to extend multi-line functionality of the library to have different prompts for different lines, like this:
```
sql> SELECT x
...> FROM y;
```
Or this:
```
zsh> python -c '
quote> print "hello"
quote> '
```

This work two complex things:
1. Make prompt a trait rather than string and pass `dyn Prompt`
2. Break highlighting between lines

The questions are:
1. Should we move prompt logic to `Helper` because it might share a tokenizer and other things with a highlighter and other parts of `Helper` trait?
2. What is a good interface for the trait, I would go with something like:
    ```
    fn get_prompt(&self, prompt_info: PromptInfo, context: Context)
    ```
    To abstract the line number, slice, whole buffer and all other things we might add in the future. But it seems the current code prefers adding that arguments explicitly on the function signature. Is that style important?
3. And how to cope with highlighting:
   * We can try to parse all the highlighting codes, reset the highlighting at the end of a line and reinsert all of them at the start of the next line after prompt
   * We can require a highlighter to close highlight line by line

    Both have some trade-offs, I'm not sure which one to choose.

Any other hints are also appreciated.

tailhook:
This commit changes multi-line input support in the following way:
* If there is no highlighter, the prompt is used in the first line and
  next lines are prepended by amount of spaces equal to the first line
  prompt
* If highlighter is used, the original prompt is passed for every line
  to the highlighter, and highlighter is free to change color or
  characters of the prompt as long as its length is the same (usual
  contract for the highlighter)

Internally this changes `layout` object to contain cursor position
without prompt size (this should make it easier to allow non-rectangular
prompts in future).

This mostly fixes #311

I've only implemented this thing for unix terminal. I'll take a look on windows support if this general approach is fine.

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -55,6 +55,10 @@ impl Highlighter for MyHelper {
         }
     }
 
+    fn has_continuation_prompt(&self) -> bool {
+        return true;
+    }
+
     fn highlight_hint<'h>(&self, hint: &'h str) -> Cow<'h, str> {
         Owned("\x1b[1m".to_owned() + hint + "\x1b[m")
     }
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -5,6 +5,7 @@ use rustyline::completion::{Completer, FilenameCompleter, Pair};
 use rustyline::config::OutputStreamType;
 use rustyline::error::ReadlineError;
 use rustyline::highlight::{Highlighter, MatchingBracketHighlighter};
+use rustyline::highlight::{PromptInfo};
 use rustyline::hint::{Hinter, HistoryHinter};
 use rustyline::{Cmd, CompletionType, Config, Context, EditMode, Editor, KeyPress};
 use rustyline_derive::{Helper, Validator};
@@ -15,6 +16,7 @@ struct MyHelper {
     highlighter: MatchingBracketHighlighter,
     hinter: HistoryHinter,
     colored_prompt: String,
+    continuation_prompt: String,
 }
 
 impl Completer for MyHelper {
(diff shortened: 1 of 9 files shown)


gwenn:
We should be able to keep the current behaviour (no continuation prompt, no space alignment).

tailhook:
Is there are good use cases for that? Or is this just for backwards compatibility?

I can do that, just not sure if it worth the complexity. In particular, how to configure that?
1. `Highlighter::has_continuation_prompt` ?
2. `Editor::set_rectrangular_prompt(true)` ?
3. `if highlighter.highlight_prompt(..).is_empty()` ?

Another question is how cursor should work? In the master, LineUp behaves as if prompt is rectangular (i.e. it keeps cursor column relative to the text, regardless of the length of the prompt).

gwenn:
As a readline alternative, we should try to keep a similar behaviour when possible.
And the screen size/resolution may be limited. Using extra space may be wasteful.
 
`Highlighter::has_continuation_prompt` seems fine.

I guess that `LineUp` and `LineDown` are the only commands that depend on the layout / number of columns. If we have a mix of long wrapped lines and short lines, we should have the same problem, no ?

tailhook:
> Highlighter::has_continuation_prompt seems fine.

Okay.

> I guess that LineUp and LineDown are the only commands that depend on the layout / number of columns.

Yes, as far as I understand for now. It will change if/when we have rectangular selection.

> If we have a mix of long wrapped lines and short lines, we should have the same problem, no ?

Yes. Vim has two different commands to handle wrapped lines `gk` and `gj` vs `k` and `j`. But implementing those requires handling line wrapping by rustyline which isn't the case now as far as I know.

----

Perhaps, it's fine to keep cursor movement as the current master, and recommend using continuation prompt in case multi-line statements are frequent enough and where column movement matters (i.e. where input is something properly indented or a table).

gwenn:
I am sorry to not be more helpful right now.

tailhook:
Done. Let me know if anything else needs to be changed.

gwenn:
I will try to have look tomorrow and keep you posted...

gwenn:
Thanks.

PR #342, opened 2020-03-06
Add `Cmd::Yield` for complex custom bindings

tailhook:
This partially replaces #293, i.e. some custom bindings can be implemented like this, although using `Yield`-style bindings is more complex so they are mostly unrelated.

The idea behind the `Yield` building is that we return from the `readline` function with a special error value which holds user's value (`Arc<dyn Any>`) along with current input. This may be used by user's REPL loop for many complex to implement but useful things like:
1. Changing prompt
2. Changing mode of operation
3. Changing Vi / Emacs mode like I've put in the `yield_binding` example as part of this PR

What do you think?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ log = "0.4"
 unicode-width = "0.1"
 unicode-segmentation = "1.0"
 memchr = "2.0"
+matches = "0.1.8"
 
 [target.'cfg(unix)'.dependencies]
 nix = "0.17"
--- a/examples/yield_binding.rs
+++ b/examples/yield_binding.rs
@@ -1,0 +1,159 @@
+use std::sync::Arc;
+use std::borrow::Cow::{self, Borrowed, Owned};
+use std::mem;
+
+use rustyline::completion::{Completer, FilenameCompleter, Pair};
+use rustyline::config::OutputStreamType;
+use rustyline::error::ReadlineError;
+use rustyline::highlight::{Highlighter, MatchingBracketHighlighter};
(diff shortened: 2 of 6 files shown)


rwestphal:
What's the status of this PR?

I'm not a maintainer but these changes look good to me.

@tailhook maybe you could move the `matches!`-related changes and the new example into separate commits to facilitate the review process.

tailhook:
> @tailhook maybe you could move the `matches!`-related changes and the new example into separate commits to facilitate the review process.

I can move them before, not after. The issue is that `Cmd` doesn't have `PartialEq` after this change, so you can only match on it. 

But generally I've not got any agreement that this functionality would be accepted. So I don't want to waste time on rebasing it until that. (And this PR is terribly outdated)

Also, the biggest issue with `Yield` being used for complex commands is that "undo" would not work well (this is still fine for other use cases, where you treat input differently if submitted by a different key binding, or if you have an external editor on a keybinding, though).

rwestphal:
> But generally I've not got any agreement that this functionality would be accepted. So I don't want to waste time on rebasing it until that. (And this PR is terribly outdated)

Fair enough. In the meanwhile I'll create a private fork of rustyline since I really need `Cmd::Yield` to implement a Cisco IOS-like CLI in Rust.

I hope this PR gets the attention it deserves as having the ability to bind custom behavior to certain keystrokes is crucial for any readline implementation.

rwestphal (approved):
LGTM

Keating950:
Is this still being considered? I'm very interested in this feature.

PR #354, opened 2020-03-28
Print an error message while trying to use the dummy implementation.

marmistrz:
It was quite confusing that rustyline did not work under WASI with no error message.

### DIFF
--- a/src/tty/test.rs
+++ b/src/tty/test.rs
@@ -159,8 +159,14 @@ impl Term for DummyTerminal {
 
     // Init checks:
 
+    #[cfg(not(target_arch = "wasm32"))]
     fn is_unsupported(&self) -> bool {
         false
+    }
+
+    #[cfg(target_arch = "wasm32")]
+    fn is_unsupported(&self) -> bool {
+        true
     }
 
     fn is_stdin_tty(&self) -> bool {


src/tty/test.rs
```diff
@@ -149,6 +149,8 @@ impl Term for DummyTerminal {
         _tab_stop: usize,
         bell_style: BellStyle,
     ) -> DummyTerminal {
+        #[cfg(not(test))]
+        log::error!("Unsupported platform, dummy terminal implementation is being used. Rustyline will not work properly");
```
bjorn3: This will only print the error if you have `RUST_LOG=error` set.

marmistrz: At least `env_logger` has error-level messages enabled by default. I tested that under WASI the message is printed without any need to set environment variables.

bjorn3: What if you don't setup a logger?

marmistrz: Then it won't. I preferred a logger, because it's less invasive.

As an alternative, we could change it to `eprintln` and possibly add a feature to silence the warning.

marmistrz:
bump

gwenn:
Sorry for the late answer.
I am not sure but there is a dedicated method ([is_unsupported](https://github.com/kkawakam/rustyline/blob/master/src/tty/mod.rs#L214)) to tell that `stdin` is bound to an unsupported terminal. When `is_unsupported` returns false, you fallback to [`io::stdin().read_line`](https://github.com/kkawakam/rustyline/blob/master/src/lib.rs#L713).
So I suggest we redefine `is_unsupported` for [DummyTerminal](https://github.com/kkawakam/rustyline/blob/master/src/tty/test.rs#L162) on wasm32 .

marmistrz:
Sorry for the late action, I misssed the notification. I updated the PR accordingly.

Issue #359, opened 2020-04-07
Windows panics with text on the same line


jonathandturner:
We're running into a situation in Nushell where we're seeing rustyline panic in Windows. It seems to happen if there is text on the same line as the prompt when it's starting.

You can reproduce this in rustyline today. In the example called 'example', change: 

https://github.com/kkawakam/rustyline/blob/master/examples/example.rs#L111

To use `print!` instead of `println!`

Here's an example of the crash:
![image](https://user-images.githubusercontent.com/547158/78724289-3c70f800-7981-11ea-80c5-47f31bac6493.png)

I can keep looking for a fix, but you all will know the codebase better and may know what could be causing that.

gwenn:
@jonathandturner Would you mind running the example with debug logs ?
```sh
> set RUST_LOG=rustyline=debug
> cargo run --example example 2> debug.log
```
And post the "initial cursor location" when the error occurs.

I tried to reproduce the crash on Windows 10 without success.

jonathandturner:
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
     Running `target\debug\examples\example.exe`
[2020-04-08T16:56:45Z DEBUG rustyline::tty::windows] ANSI colors already enabled
[2020-04-08T16:56:45Z DEBUG rustyline::edit] old layout: Layout { prompt_size: Position { col: 0, row: 0 }, default_prompt: false, cursor: Position { col:
 0, row: 0 }, end: Position { col: 0, row: 0 } }
[2020-04-08T16:56:45Z DEBUG rustyline::edit] new layout: Layout { prompt_size: Position { col: 3, row: 0 }, default_prompt: true, cursor: Position { col:
3, row: 0 }, end: Position { col: 3, row: 0 } }
1> [2020-04-08T16:56:47Z DEBUG rustyline::keymap] Emacs command: AcceptLine
[2020-04-08T16:56:47Z DEBUG rustyline::undo] Changeset::begin
[2020-04-08T16:56:47Z DEBUG rustyline::undo] Changeset::end

[2020-04-08T16:56:47Z DEBUG rustyline::tty::windows] ANSI colors already enabled
Line: [2020-04-08T16:56:47Z DEBUG rustyline::tty::windows] initial cursor location: 6, 38

Error: Io(Os { code: 87, kind: Other, message: "The parameter is incorrect." })
```

I noticed that you might need to be in Windows Terminal and not the old cmd to repro the issue.

[...]

PR #361, opened 2020-04-09
Try to fix move_cursor_at_leftmost on Windows terminal

gwenn:
Only tested on Wine.
TODO test on Windows.


gwenn:
Like PR #360
But keep `set_console_cursor_position` untouched.
Only perform `println!("")` while trying to move cursor at leftmost.

jonathandturner:
I've just tested this PR on my setup, and this doesn't fix #359.

Issue #366, opened 2020-04-18
Does clearing screen reset background also


jonathandturner:
Recently, we had someone file this issue for nushell: https://github.com/nushell/nushell/issues/1599

It seems as though when they use the clear screen shortcut, rustyline is able to clear the screen. But, when there's a background color (like the red we use for errors), that doesn't get cleared.

Is there a way to enable clearing the background also?

bjorn3:
Unix uses `\x1b[H\x1b[2J` which clears the background:

https://github.com/kkawakam/rustyline/blob/cbedcf1df1905840deae4463bf060e6015c341f6/src/tty/unix.rs#L656

Windows uses `FillConsoleOutputCharacterA` to fill the screen with spaces. It seems that this doesn't affect colors. According to https://docs.microsoft.com/en-us/windows/console/clearing-the-screen you need to call `FillConsoleOutputAttribute` as well.

https://github.com/kkawakam/rustyline/blob/d95cd7c75a0111012e387612f885165e8abf0498/src/tty/windows.rs#L292-L298

gwenn:
Indeed, we should do both like
- [Haskeline](https://github.com/judah/haskeline/blob/master/System/Console/Haskeline/Backend/Win32.hsc#L562)
- [Linefeed](https://github.com/murarth/mortal/blob/master/src/windows/terminal.rs#L880)
- [python-prompt-toolkit](https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/prompt_toolkit/output/win32.py#L257)

gwenn:
Could you please give a try to PR #367 ?
Thanks.

jonathandturner:
Yes, this looks better. I couldn't get it to repro for the things I tried.

jonathandturner:
@gwenn - can we merge this fix for the next release, too?

gwenn:
@jonathandturner 6.1.2 has been published.

PR #367, opened 2020-04-19
Fix clear_screen on Windows

gwenn:
We should both clear content and attributes.
See #366

### DIFF
--- a/src/tty/windows.rs
+++ b/src/tty/windows.rs
@@ -287,7 +287,7 @@ impl ConsoleRenderer {
         Ok(())
     }
 
-    fn clear(&mut self, length: DWORD, pos: wincon::COORD) -> Result<()> {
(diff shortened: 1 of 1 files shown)


src/tty/windows.rs
```diff
@@ -296,6 +296,13 @@ impl ConsoleRenderer {
             pos,
             &mut _count,
         ));
+        check!(wincon::FillConsoleOutputAttribute(
+            self.handle,
+            attr,
+            length,
+            pos,
+            &mut _count
```
bjorn3, line 304:
```suggestion
            &mut _count,
```

gwenn: `cargo fmt` doesn't add a comma !

bjorn3: That is because this is inside a macro, so it doesn't know it is allowed to do that.

gwenn: Indeed !

Issue #208, opened 2019-04-08
vi modal indicator


shanemikel:
Readline supports the vi editing mode, with these options in `~/.inputrc`
```
set editing-mode vi
set show-mode-in-prompt on
set vi-cmd-mode-string :
set vi-ins-mode-string +
```

The `show-mode-in-prompt` feature is a really helpful bit of user feedback for a modal editing system like Vi.  You can test this feature by writing the configuration above, and running Bash.

AdminXVII:
I think a good way would be to follow NVim's way: using a line cursor ('\033[6 q') for insert and emacs mode and switch to a block cursor for command mode.

shanemikel:
I disagree @AdminXVII.  The issue is that not all terminals support cursor modes.

The default behavior of iPython (which doesn't use readline anymore) is to show `[ins]` in insert mode, and `[nav]` in normal/navigation mode.  I think that's pretty good and self explanatory.

shanemikel:
I should also say that the default indicators of readline are `(cmd)` for normal/command mode and `(ins)` for insert mode.  Maybe it's better to use the readline defaults given the library's lineage.

max-sixty:
FWIW this is the config I use for IPython, to get cursors:

```python
   3   │ import sys
   4   │ from prompt_toolkit.key_binding.vi_state import InputMode, ViState
   5   │
   6   │
   7   │ def get_input_mode(self):
   8   │     return self._input_mode
   9   │
  10   │
  11   │ def set_input_mode(self, mode):
  12   │     shape = {InputMode.NAVIGATION: 2, InputMode.REPLACE: 4}.get(mode, 6)
  13   │     raw = u'\x1b[{} q'.format(shape)
  14   │     if hasattr(sys.stdout, '_cli'):
  15   │         out = sys.stdout._cli.output.write_raw
  16   │     else:
  17   │         out = sys.stdout.write
  18   │     out(raw)
  19   │     sys.stdout.flush()
  20   │     self._input_mode = mode
  21   │
  22   │
  23   │ ViState._input_mode = InputMode.INSERT
  24   │ ViState.input_mode = property(get_input_mode, set_input_mode)
  25   │ c.TerminalInteractiveShell.editing_mode = 'vi'
  26   │ c.TerminalInteractiveShell.prompt_includes_vi_mode = False
```

pjetherton:
> The issue is that not all terminals support cursor modes.

[...]

PR #369, opened 2020-04-19
Vi mode indicators

elshize:
My first take on  #208

I personally agree with @AdminXVII that it would be nice to have a line cursor, maybe behind an option if there are support concerns. But I wasn't sure how to do that, so I started off with this.

I realize it's not finished yet (for one, I haven't included any unit test that work with the indicators) but I figured it would be nice to get some feedback to see if I'm going the right direction.

### DIFF
--- a/examples/vi.rs
+++ b/examples/vi.rs
@@ -6,7 +6,7 @@ use rustyline::error::ReadlineError;
 use rustyline::highlight::{Highlighter, MatchingBracketHighlighter, PromptState};
 use rustyline::hint::{Hinter, HistoryHinter};
 use rustyline::validate::{self, MatchingBracketValidator, Validator};
-use rustyline::{Cmd, CompletionType, Config, Context, EditMode, Editor, KeyPress};
+use rustyline::{Cmd, CompletionType, Config, Context, EditMode, Editor, InputMode, KeyPress};
 use rustyline_derive::Helper;
 
 #[derive(Helper)]
@@ -15,7 +15,6 @@ struct MyHelper {
     highlighter: MatchingBracketHighlighter,
     validator: MatchingBracketValidator,
     hinter: HistoryHinter,
-    colored_prompt: String,
 }
 
 impl Completer for MyHelper {
@@ -44,7 +43,13 @@ impl Highlighter for MyHelper {
         prompt_state: PromptState,
     ) -> Cow<'b, str> {
         if prompt_state.default {
-            Borrowed(&self.colored_prompt)
+            let indicator = match prompt_state.mode {
+                Some(InputMode::Command) => ":",
+                Some(InputMode::Insert) => "+",
+                Some(InputMode::Replace) => "#",
+                _ => " ",
+            };
+            Owned(format!("\x1b[1;32m{}{}\x1b[0m", indicator, &prompt[1..]))
         } else {
             Borrowed(prompt)
         }
@@ -83,14 +88,13 @@ fn main() -> rustyline::Result<()> {
     let config = Config::builder()
         .history_ignore_space(true)
         .completion_type(CompletionType::List)
-        .edit_mode(EditMode::Emacs)
+        .edit_mode(EditMode::Vi)
         .output_stream(OutputStreamType::Stdout)
         .build();
     let h = MyHelper {
         completer: FilenameCompleter::new(),
         highlighter: MatchingBracketHighlighter::new(),
         hinter: HistoryHinter {},
-        colored_prompt: "".to_owned(),
         validator: MatchingBracketValidator::new(),
     };
     let mut rl = Editor::with_config(config);
@@ -102,8 +106,7 @@ fn main() -> rustyline::Result<()> {
     }
     let mut count = 1;
     loop {
-        let p = format!("{}> ", count);
-        rl.helper_mut().expect("No helper").colored_prompt = format!("\x1b[1;32m{}\x1b[0m", p);
+        let p = format!(" {}> ", count);
         let readline = rl.readline(&p);
         match readline {
             Ok(line) => {
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,7 +2,7 @@
 use std::default::Default;
 
 /// User preferences
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Config {
     /// Maximum number of entries in History.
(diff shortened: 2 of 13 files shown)


gwenn:
I am not sure it is feasible but I would prefer a less intrusive change:
* add a dedicated placeholder in your default (not highlighted) prompt. (You need this because the prompt size must not change after highlighting)
* refactor [highlight_prompt](https://docs.rs/rustyline/6.1.1/rustyline/highlight/trait.Highlighter.html#method.highlight_prompt) such as you have access to current editing mode. If possible, try to introduce an indirection (like `PromptState`) because recently a similar change was introduced (see https://github.com/kkawakam/rustyline/pull/338/files#diff-3f1033c6837e548c84459c1469e59a4bR38):
```rust
pub struct PromptState<'a> {
     pub default: bool,
     pub mode: InputMode,
 }

pub trait Highlighter {
...
    fn highlight_prompt<'b, 's: 'b, 'p: 'b>(
        &'s self,
        prompt: &'p str,
        state: PromptState,
    ) -> Cow<'b, str> {
...
```
* reuse/keep the changes you already made to refresh the line when the mode changes.
* implement a ViModeHighlighter either in a dedicated example or as a reusable component like `MatchingBracketHighlighter` by replacing the placeholder with indicator matching the current editing mode.

I may investigate further if I have time.
Thanks.

elshize:
I see. I think this is quite doable. My only concern would be that it essentially would mean that all your indicators must be the same width. This could be a bit limiting in general (I see no reason why someone shouldn't be able to use `- INSERT -` and `- COMMAND -` if they want to). The solution could be (a) we don't care, that's all we allow, or (b) we don't care right now, but will reconsider later.

Regarding (b), is there a plan to allow for highlighting to have a different width or is it here to stay?

elshize:
Is this one more like what you were thinking? (Probably want a separate example for Vi mode, but wanted to illustrate quickly the approach.)

gwenn:
The limitation `width(raw prompt) = width(highlighted prompt)` is because the repaint/refresh phase/code is not clever nor fast.

src/tty/mod.rs
```diff
@@ -39,6 +40,7 @@ pub trait Renderer {
         old_layout: &Layout,
         new_layout: &Layout,
         highlighter: Option<&dyn Highlighter>,
+        vi_mode: Option<InputMode>,
```
gwenn, line 43: You can factorize code by passing a `PromptState` instead of `Option<InputMode>`.

elshize:
About the trait, are you thinking of having different implementations for vi/emacs?

elshize:
Also, I wasn't sure what your preference on example is, whether extract the common code or maybe use a more minimal example for Vi mode? Or maybe parameterize the default example with input mode?

gwenn:
I would prefer you keep the default example untouched.

gwenn:
Oops!

gwenn:
> About the trait, are you thinking of having different implementations for vi/emacs?
I am not sure but it may be easier to extend the trait in the future and we already have `State` as implementation.

examples/vi.rs
```diff
@@ -0,0 +1,132 @@
+use std::borrow::Cow::{self, Borrowed, Owned};
```
gwenn, line 1: You should have keep your example minimalist (like https://github.com/kkawakam/rustyline/blob/master/examples/read_password.rs)

src/config.rs
```diff
@@ -2,7 +2,7 @@
 use std::default::Default;
 
 /// User preferences
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
```
gwenn, line 5: To be reverted.

src/edit.rs
```diff
@@ -33,6 +34,7 @@ pub struct State<'out, 'prompt, H: Helper> {
     pub ctx: Context<'out>,   // Give access to history for `hinter`
     pub hint: Option<String>, // last hint displayed
     highlight_char: bool,     // `true` if a char has been highlighted
+    input_mode: Option<InputMode>,
```
gwenn, line 37: We may avoid to duplicate input_mode if we move `InputState` in `State` ?

src/history.rs
```diff
@@ -319,7 +319,7 @@ mod tests {
     #[test]
     fn add() {
         let config = Config::builder().history_ignore_space(true).build();
-        let mut history = History::with_config(config);
+        let mut history = History::with_config(config.clone());
```
gwenn, line 322: To be reverted.

src/lib.rs
```diff
@@ -784,7 +784,7 @@ impl<H: Helper> Editor<H> {
         );
         Self {
             term,
-            history: History::with_config(config),
+            history: History::with_config(config.clone()),
```
gwenn, line 787: To be reverted.

src/config.rs
elshize: :+1: missed that.

elshize:
Thanks for the feedback, I'll try to push some changes soon.

elshize:
The problem with using `State` as implementation of `PromptState` is that we'd have to pass it to `refresh` by reference while the writer that is part of the state is passed by a mutable reference at the same time. So I'm not sure if it's worth it at this point. I could extract `default_prompt` from layout and put into an object together with `input_mode` inside a `PromptState` implementation but I don't think it's worth the hassle right now.

What do you think?

gwenn:
We should make `PromptState` compatible with #372.

elshize:
Ok, so extracting `Prompt` helps with the issue I raised (`Prompt` could implement `PromptState` trait and be passed) but that commit also introduces `PromptInfo`, which has eve more data in it. So if you want it to be compatible, I think the easiest way would be to add `Option<InputMode>` to `PromptInfo`.

Either way, do you want me to wait for that other PR to be merged and work on top of that? It might be easier for everyone.

gwenn:
I am afraid there will be no indicator when colors are disabled (`ColorMode::Disabled`) or not supported (Windows < 10).

gwenn:
There is a draft [here](https://github.com/kkawakam/rustyline/compare/master...gwenn:prompt#diff-d7f61085d308113c354cf49ee24427df) to support dynamic prompt even if there is no highlighting.
First, I will try to make it build/work with current master branch: only static prompt.
Then I will try to integrate your changes.

elshize:
Thanks for the update, let me know if I can help with anything.

PR #370, opened 2020-04-21
Hide cursor while refreshing line on Windows

### LINKED ISSUE #256: Flicker at first character of the line in Windows
Rustyline currently seems to flicker at the first character when typing/backspacing, but this seems to only affect Windows - I haven't seen it on Linux.

I tried to record an animated gif so you can see it:

rustyline_flicker

In practice, it flickers more than you can see in the gif, but this should at least show where it's flickering.

gwenn:
~~TODO scopeguard.~~
Should fix #256.


jonathandturner:
@gwenn - I don't see any flicker anymore with this fix

PR #376, opened 2020-05-04
Simplify syntax

timClicks:
Some minor tweaks:

- Remove explicit returns
- Removes redundant tokens, e.g. `ValidationResult`, `Ok`
- Puts all `ValidationResult`s on the same indentation level


gwenn:
Thanks.

PR #379, opened 2020-05-08
Flexible AcceptLine, fix bug in multi-line accept

tailhook:
1. This commit allows keybinding for forcing accept line (even if
  validator fails). This is useful if you have a separate keybinding
  for accepting, such as `Alt-Enter`. It makes sense to show an error
  to user rather than insert an unexpected newline.
2. Forcing `Newline` is also accepted
3. `AcceptLine` meaning is changed (now it forces accept),
   but default `Enter` keybinding is not
4. `AcceptOrInsertLine` now has a flag that modifies it's behavior in
  the middle of the line (at the end it always submits, when validator
  returned false it's always a newline)
5. Fixed bug with exiting readline (both on success and error) when
  cursor is not on the last line

I'm not sure that current command-names are the best possible, so let me know if we can do better. The primary motivator was (1) from above.

This conflicts with #342. I'll rebase other one when either of them is merged.

### DIFF
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -20,6 +20,8 @@ pub enum Cmd {
     /// abort
     Abort, // Miscellaneous Command
     /// accept-line
+    ///
+    /// See also AcceptOrInsertLine
     AcceptLine,
     /// beginning-of-history
     BeginningOfHistory,
@@ -99,9 +101,23 @@ pub enum Cmd {
     /// moves cursor to the line below or switches to next history entry if
     /// the cursor is already on the last line
     LineDownOrNextHistory(RepeatCount),
-    /// accepts the line when cursor is at the end of the text (non including
-    /// trailing whitespace), inserts newline character otherwise
-    AcceptOrInsertLine,
+    /// Inserts a newline
+    Newline,
+    /// Either accepts or inserts a newline
+    ///
+    /// Always inserts newline if input is non-valid. Can also insert newline
+    /// if cursor is in the middle of the text
+    ///
+    /// If you support multi-line input:
+    /// * Use `accept_in_the_middle: true` for mostly single-line cases,
+    ///   for example command-line.
+    /// * Use `accept_in_the_middle: false` for mostly multi-line cases,
+    ///   for example SQL or JSON input.
+    AcceptOrInsertLine {
+        /// Whether this commands accepts input if the cursor not at the end
+        /// of the current input
+        accept_in_the_middle: bool
+    },
 }
 
 impl Cmd {
@@ -913,7 +929,9 @@ impl InputState {
                 }
             }
             KeyPress::Ctrl('J') |
-            KeyPress::Enter => Cmd::AcceptLine,
+            KeyPress::Enter => {
+                Cmd::AcceptOrInsertLine { accept_in_the_middle: true }
+            }
             KeyPress::Down => Cmd::LineDownOrNextHistory(1),
             KeyPress::Up => Cmd::LineUpOrPreviousHistory(1),
             KeyPress::Ctrl('R') => Cmd::ReverseSearchHistory,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -282,7 +282,7 @@ fn page_completions<C: Candidate, H: Helper>(
                 && cmd != Cmd::Kill(Movement::BackwardChar(1))
                 && cmd != Cmd::AcceptLine
                 && cmd != Cmd::Newline
-                && matches!(cmd, Cmd::AcceptOrInsertLine { .. })
+                && !matches!(cmd, Cmd::AcceptOrInsertLine { .. })
             {
                 cmd = s.next_cmd(input_state, rdr, false)?;
             }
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -508,9 +510,12 @@ fn readline_edit<H: Helper>(
             }
             Cmd::EndOfFile => {
                 if !input_state.is_emacs_mode() && !s.line.is_empty() {
-                    s.edit_move_end()?;
                     break;
                 } else if s.line.is_empty() {
+                    // Move to end, in case cursor was in the middle of the
+                    // line, so that next thing application prints goes after
+                    // the input
+                    s.edit_move_buffer_end()?;
```
gwenn, line 518: If `line` is empty, there is no need to move the cursor, no?

tailhook: Yes. But we need to clear the hint. Fixed.

src/lib.rs
```diff
@@ -281,15 +281,16 @@ fn page_completions<C: Candidate, H: Helper>(
                 && cmd != Cmd::SelfInsert(1, ' ')
                 && cmd != Cmd::Kill(Movement::BackwardChar(1))
                 && cmd != Cmd::AcceptLine
-                && cmd != Cmd::AcceptOrInsertLine
+                && cmd != Cmd::Newline
+                && matches!(cmd, Cmd::AcceptOrInsertLine { .. })
```
gwenn, line 285: `&& !matches!(cmd, Cmd::AcceptOrInsertLine { .. })` (not matches)

src/lib.rs
```diff
@@ -509,8 +511,12 @@ fn readline_edit<H: Helper>(
                 s.edit_overwrite_char(c)?;
             }
             Cmd::EndOfFile => {
+                if s.has_hint() || !s.is_default_prompt() {
```
gwenn, line 514: Only if line is empty or in vi mode ?
That is when the action is not `edit_delete`.

src/lib.rs
```diff
@@ -585,13 +591,23 @@ fn readline_edit<H: Helper>(
...
+                let valid = s.validate()?;
+                let end = s.line.is_end_of_input();
+                match (cmd, valid, end) {
+                    | (Cmd::AcceptLine, _, _)
+                    | (Cmd::AcceptOrInsertLine { .. }, true, true)
+                    | (Cmd::AcceptOrInsertLine { accept_in_the_middle: true }, true, _)
+                    => {
+                        break;
+                    }
+                    | (Cmd::Newline, _, _)
```
gwenn, line 603: Unreachable ?

src/lib.rs
```diff
@@ -656,6 +672,10 @@ fn readline_edit<H: Helper>(
                 }
             }
             Cmd::Interrupt => {
+                // Move to end, in case cursor was in the middle of the
+                // line, so that next thing application prints goes after
+                // the input
+                s.edit_move_end()?;
```
gwenn, line 678: edit_move_buffer_end ?

src/keymap.rs
```diff
@@ -913,7 +929,9 @@ impl InputState {
                 }
             }
             KeyPress::Ctrl('J') |
-            KeyPress::Enter => Cmd::AcceptLine,
+            KeyPress::Enter => {
```
gwenn, line 932: `AcceptLine == AcceptOrInsertLine { accept_in_the_middle: true }` ?

src/lib.rs
tailhook: Fixed

src/lib.rs
tailhook: Fixed. But I'm not sure that `if !input_state.is_emacs_mode() && !s.line.is_empty()` makes sense. `Ctrl+D` should be "scroll downwards" in vim, not "submit".

src/keymap.rs
tailhook: Yes, accept whatever is fully formed command or insert a newline otherwies. That's for backwards compatibility.  As described in the docstring above this is what makes most sense for mostly-single-line cases (i.e. a command-line like in nushell).

src/lib.rs
tailhook: Fixed. Thanks!

(this is actually why I'd like the top-level match to be exhaustive, as this error would have been spotted by the compiler)

src/lib.rs
tailhook: Yes, that's better. Thanks!

gwenn:
Thanks you.
I will try to review your PR as soon as possible.

src/lib.rs
```diff
@@ -509,13 +511,19 @@ fn readline_edit<H: Helper>(
...
                     s.edit_delete(1)?
+                } else {
+                    if s.has_hint() || !s.is_default_prompt() {
+                        // Force a refresh without hints to leave the previous
+                        // line as the user typed it after a newline.
+                        s.refresh_line_with_msg(None)?;
+                    }
+                    if s.line.is_empty() {
+                        return Err(error::ReadlineError::Eof);
+                    } else if !input_state.is_emacs_mode() {
```
gwenn, line 524: Always `true`

src/lib.rs
```diff
@@ -509,13 +511,19 @@ fn readline_edit<H: Helper>(
...
+                        break;
```
gwenn, line 525: `s.edit_move_end()?;` + `break;` like before ?

Issue #382, opened 2020-05-15
Windows tab completion of files and dirs


fdncred:
I was looking at the completion code trying to figure out what would need to change so that it works better on Windows. Rustyline seems to do well with case sensitive completions but not as well on Windows where case insensitive is needed.

I was thinking this may be where the code needs to change but I could be reading this all wrong.
https://github.com/kkawakam/rustyline/blob/e7bf9012197717bbaad5f75ed91583886d906038/src/completion.rs#L338

As an example, I'd like to autocomplete `C:\Apps` by typing `cd c:\ap<tab>` but since it's case sensitive. Or vice-versa, if my folder is called `C:\apps` and I type `CD C:\A<tab>` I'd like it to autocomplete.

Is changing to case insensitive an option via configuration somehow that I have missed perhaps? Or is the only way with a PR?

gwenn:
No, there is no option to make completion case-insensitive.

fdncred:
Is that line the appropriate place to make a case insensitive change?

gwenn:
I think so.
At least `macos` and `windows` are case insensitive.

fdncred:
Wait a second, I'm primarily a user on `windows` but also use it on `macos`. I was considering a PR because it is case sensitive in Windows. MacOs works great. Are you saying that Windows implementations should already be case insensitive?

gwenn:
I should have said:
At least `macos` and `windows` _should be_ case insensitive.

PR #387, opened 2020-05-18
Allow Windows to use case insensitive path completion

fdncred:


fdncred:
closes #382

gwenn:
See #389.
Just to avoid code duplication.
And to make it works on windows and macos.

fdncred:
Agreed. Much better.

PR #388, opened 2020-05-19
Migrate to dirs-next

gwenn:


decathorpe:
Hm. It looks like the original dirs crate isn't abandoned after all. Since there were two new releases in the past month, it's been updated more recently than the dirs-next fork - so I'm unsure whether it's a good idea to stick with the fork or switch back to the original crate :frowning:

PR #389, opened 2020-05-20
Allow Windows to use case insensitive path completion

### LINKED ISSUE #382: Windows tab completion of files and dirs

gwenn:
Fix #382

### DIFF
--- a/src/completion.rs
+++ b/src/completion.rs
@@ -344,10 +344,12 @@ fn filename_complete(
 
     // if any of the below IO operations have errors, just ignore them
     if let Ok(read_dir) = dir.read_dir() {
+        let file_name = normalize(file_name);
         for entry in read_dir {
             if let Ok(entry) = entry {
                 if let Some(s) = entry.file_name().to_str() {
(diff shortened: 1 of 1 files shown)


fdncred:
@gwenn if you don't mind my asking, when is the next release? anxious to get this change in.

PR #403, opened 2020-06-17
Avoid EINTR in readline by repeating poll-call.

### LINKED ISSUE #402: Avoid returning EINTR in PosixRenderer::move_cursor_at_leftmost
Hello,
I could trace back a weird termination in nushell on linux to an EINTR returned by a `poll` call in rustyline.

The issue occurs when I run `qterminal -e nu`, but is timing dependent and may be not reproducible on other devices. Even small changes like additional logging changed the behavior during debugging.

There are multiple potential fixes:

1. Repeat the call on EINTR.
2. use `sigaction` to repeat all interrupted syscalls. This may prevent multiple similar issues, but unfortunately it is a program-wide change.

I will send a pull request soon, with a fix for option 1.

Marthog:
Fixes #402 by repeating the call on EINTR.

### DIFF
--- a/src/tty/unix.rs
+++ b/src/tty/unix.rs
@@ -390,8 +390,16 @@ impl PosixRawReader {
     }
 
     fn poll(&mut self, timeout_ms: i32) -> ::nix::Result<i32> {
+        use nix::{Error, errno::Errno};
+
         let mut fds = [poll::PollFd::new(STDIN_FILENO, PollFlags::POLLIN)];
-        poll::poll(&mut fds, timeout_ms)
+        loop {
+            let polled = poll::poll(&mut fds, timeout_ms);
+            match polled {
+                Err(Error::Sys(Errno::EINTR)) => continue,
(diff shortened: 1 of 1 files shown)


gwenn:
Thanks.
But I am not sure we should loop.
* https://github.com/kkawakam/rustyline/blob/master/src/tty/unix.rs#L417
Here, we try to make the distinction between a single ESC and an escape sequence.
I guess we can safely assume that a single ESC was received if an EINTR is raised.
* https://github.com/kkawakam/rustyline/blob/master/src/tty/unix.rs#L705
Here, we try to find the current cursor position.
I guess we can safely assume that cursor position is OK if an EINTR is raised.

bjorn3:
An `EINTR` can be caused by any signal, including `SIGSTOP`, `SIGCHLD`, `SIGWINCH` or `SIGUSR1`.

gwenn:
See #420.

Issue #409, opened 2020-06-29
Windows and paths with n in them


fdncred:
Windows 10 2004.

Somewhere along the way, paths are getting stored in the history.txt file wrong. For instance, I work on a project called NuShell. Here's an entry from my history.txt file.
`
cd source\\repos\\forks\\\nushell\\
`
Notice that there are one too many back slashes in the entry. The ramification of this is that when I try to recall it, this happens.
![image](https://user-images.githubusercontent.com/343840/86010341-e3918880-b9e0-11ea-8adc-22c6a921ddbf.png)

Any ideas what is going on here? I'm assuming it has something to do with either the history loading or saving or both.

bjorn3:
This is expected. Both backslash and newline are escaped when writing the history file. They are unescaped when loading it again. The history file has a single history entry per line, so newlines must be escaped as `\n`. This in turn requires `\` to be escaped as `\\` to prevent ambiguity when you write `\n` and not a newline.

https://github.com/kkawakam/rustyline/blob/380995919c0dc0565aa71b352400ed0e5f4a15d0/src/history.rs#L143-L144

https://github.com/kkawakam/rustyline/blob/380995919c0dc0565aa71b352400ed0e5f4a15d0/src/history.rs#L173

fdncred:
> This is expected. 

So, it's expected to not work properly with any path that has a folder that begins with `n`?

What about checking for `\n` at the end of a string instead of in the middle. Would that work?

bjorn3:
> Notice that there are one too many back slashes in the entry.

I didn't notice this part when I responded. This is not expected. I think the problem here is that `\n` got turned into `\\n` when written to the history file. So far so good, but when it got read again, the `\n` part first got turned into a newline to form `\<LF>`, this is not correct. When this is written again it forms `\\\n`. The solution would be to replace the two `replace` calls with a single pass that replaces `"\\n"` with `"\n"` and `"\\\\"` with `"\\"` at the same time.

gwenn:
Thanks you for both report and analysis.

gwenn:
See #412
There may be a more idiomatic way to lazily / conditionally make a partial copy (https://github.com/kkawakam/rustyline/pull/412/files#diff-fe6f8eb00378a6fd24f0b8c7f0cb9cebR194) ?

gwenn:
6.3.0 released

Issue #400, opened 2020-06-15
history left in broken state


avsej:
https://github.com/nushell/nushell/issues/1977

The library always overwrites the history which does not have practical value. Instead it must only append new lines to the existing file.

gwenn:
https://github.com/kkawakam/rustyline/blob/master/src/history.rs#L125
```rust
    // TODO append_history
```
not supported.

gwenn:
Could you please give a try to PR #419 ?

gwenn:
Version 7.0.0 released

PR #411, opened 2020-06-29
Updated history to escape history entries differently

fdncred:
@gwenn How's this for a fix for #409?

I used the  `snailquote` escaping/unescaping crate found [here](https://docs.rs/snailquote/0.3.0/snailquote/)

I tested it with nushell and it seemed to work, once I straightened out my history.txt file

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ log = "0.4"
 unicode-width = "0.1"
 unicode-segmentation = "1.0"
 memchr = "2.0"
+snailquote = "0.3.0"
 
 [target.'cfg(unix)'.dependencies]
 nix = "0.17"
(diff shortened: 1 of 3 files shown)


fdncred:
There is an `unwrap()` in there because I wasn't sure what else to do.

gwenn:
Thanks for your PR but:
* The only thing we want is to make the distinction between single-line and multi-lines entry in history.
To do so we need to escape '\n'.
But we have to also escape '\\' to avoid ambiguity.
And that's all !
* We may also need to timestamp entries for #400.
So it seems wise to implement our own "parser" which should help supporting new format(s).

See #412.

fdncred:
@gwenn Ok. Looks good to me. Thanks for putting my test in there.

PR #412, opened 2020-06-30
Fix History::load

gwenn:
* Improve History::save
* Add unit test to reproduce bug #409
* Fix History::load


fdncred:
Great! it'll be nice to have this in our next release. Thanks @gwenn!

Issue #413, opened 2020-07-07
Support partial completions


ryanpcmcquen:
Would it be possible to get partial completions, like `fish` provides with the <kbd>Alt</kbd> and arrow combos?

Example:
![partial_completion](https://user-images.githubusercontent.com/772937/86825136-dfe6ad00-c043-11ea-9751-93b12d80ffb0.gif)

I also logged the request here as well:
https://github.com/nushell/nushell/issues/2135

gwenn:
Version 8.1.0 released

PR #414, opened 2020-07-09
Detach hint title from completion text

### LINKED ISSUE #397: What is the purposes of hint?
I'm using hint to hints of the next arguments of the command, like this:
```
some-flag [on|off]
         ^ cursor is here
```
Or even for descriptions of the commands:
```
set -- set configuration value
   ^ cursor is here
```
But in this case `Right` arrow at the end of input puts the hint into the input line. Which is not intended. `Tab` does completion as intended
(And as far as I understand I can't override right arrow to `Complete` because the right arrow in the middle of the input will not work as cursor movement).

Is it wrong usage of hint or is it a bug?

tailhook:
This solution came up in disussion of #397

Fixes #397, and perhaps #413

### DIFF
--- a/rustyline-derive/src/lib.rs
+++ b/rustyline-derive/src/lib.rs
@@ -50,6 +50,12 @@ pub fn hinter_macro_derive(input: TokenStream) -> TokenStream {
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
     let expanded = quote! {
         impl #impl_generics rustyline::hint::Hinter for #name #ty_generics #where_clause {
+            type Hint = String;
+            fn hint(&self, _line: &str, _pos: usize, _ctx: &rustyline::Context<'_>)
+                -> Option<Self::Hint>
+            {
+                None
+            }
         }
     };
     TokenStream::from(expanded)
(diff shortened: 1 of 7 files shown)


rustyline-derive/src/lib.rs
```diff
@@ -50,6 +50,12 @@ pub fn hinter_macro_derive(input: TokenStream) -> TokenStream {
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
     let expanded = quote! {
         impl #impl_generics rustyline::hint::Hinter for #name #ty_generics #where_clause {
+            type Hint = String;
+            fn hint(&self, _line: &str, _pos: usize, _ctx: &rustyline::Context<'_>)
+                -> Option<Self::Hint>
+            {
+                None
+            }
```
gwenn, line 58: Useless !

rustyline-derive/src/lib.rs
```diff
@@ -50,6 +50,12 @@ pub fn hinter_macro_derive(input: TokenStream) -> TokenStream {
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
     let expanded = quote! {
         impl #impl_generics rustyline::hint::Hinter for #name #ty_generics #where_clause {
+            type Hint = String;
```
gwenn, line 53: `std::string::String`

tailhook:
Thank you. Sorry that I didn't find time to finish this.

PR #419, opened 2020-07-25
Append histo

### LINKED ISSUE #400: history left in broken state

gwenn:
Breaking change: a mutable reference is now needed to save history
Should fix #400

### DIFF
--- a/src/history.rs
+++ b/src/history.rs
@@ -99,7 +99,7 @@ impl History {
             self.entries.pop_front();
         }
         self.entries.push_back(line.into());
-        self.new_entries = (self.new_entries + 1).min(self.len());
+        self.new_entries = self.new_entries.saturating_add(1).min(self.len());
         true
     }
 
@@ -150,7 +150,7 @@ impl History {
         fix_perm(file);
         let mut wtr = BufWriter::new(file);
         let first_new_entry = if append {
-            self.entries.len() - self.new_entries
+            self.entries.len().saturating_sub(self.new_entries)
         } else {
             wtr.write_all(Self::FILE_VERSION_V2.as_bytes())?;
             0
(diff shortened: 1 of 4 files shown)


src/history.rs
```diff
@@ -158,6 +153,40 @@ impl History {
...
+        file.lock_exclusive()?;
+        // we may need to truncate file before appending new entries
+        let mut other = Self {
+            entries: VecDeque::new(),
+            max_len: self.max_len,
+            ignore_space: self.ignore_space,
+            ignore_dups: self.ignore_dups,
+            new_entries: 0,
+        };
+        other.load(path)?;
```
gwenn, line 182: I am not sure we can open `path` in `load` after `file.lock_exclusive`.

src/history.rs
```diff
@@ -158,6 +153,40 @@ impl History {
...
+        let first_new_entry = self.entries.len() - self.new_entries;
+        for entry in self.entries.iter().skip(first_new_entry) {
+            other.add(entry);
+        }
+        other.save(path)?;
```
gwenn, line 187: I am not sure we can open `path` in `save` after `file.lock_exclusive()`.

src/history.rs
gwenn: Ok, it works only on unix (or at least linux) but not on windows (see error in CI)...

src/history.rs
```diff
@@ -158,6 +172,48 @@ impl History {
...
+        use fs2::FileExt;
+        use std::io::Seek;
+
+        if self.is_empty() || self.new_entries == 0 {
+            return Ok(());
+        }
+        let path = path.as_ref();
+        if !path.exists() || self.new_entries == self.max_len {
+            return self.save(path);
+        } else if self.can_just_append(path)? {
```
gwenn, line 191: Maybe we should lock file before calling `can_just_append` ?

PR #422, opened 2020-07-28
Changes to allow Windows backspace keybindings

fdncred:
Added the ability to use `Ctrl+Backspace`, `Meta+Backspace`, and `Shift+Backspace` as keybindings on Windows. Thanks for your help @gwenn. I figure you'll have changes and not like some of the things I've done. :) I didn't run `cargo fmt --all` because I'm not sure if that's your practice.

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -83,7 +83,7 @@ fn main() -> rustyline::Result<()> {
     let config = Config::builder()
         .history_ignore_space(true)
         .completion_type(CompletionType::List)
-        .edit_mode(EditMode::Vi)
+        .edit_mode(EditMode::Emacs)
         .output_stream(OutputStreamType::Stdout)
         .build();
     let h = MyHelper {
@@ -97,9 +97,6 @@ fn main() -> rustyline::Result<()> {
     rl.set_helper(Some(h));
     rl.bind_sequence(KeyPress::Meta('N'), Cmd::HistorySearchForward);
     rl.bind_sequence(KeyPress::Meta('P'), Cmd::HistorySearchBackward);
-    rl.bind_sequence(KeyPress::ControlBackspace, Cmd::Kill(Movement::BackwardWord(1, Word::Vi)));
-    rl.bind_sequence(KeyPress::MetaBackspace, Cmd::Kill(Movement::BackwardChar(1)));
-    rl.bind_sequence(KeyPress::ShiftBackspace, Cmd::Kill(Movement::BackwardChar(1)));
     if rl.load_history("history.txt").is_err() {
         println!("No previous history.");
     }
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -6,7 +6,7 @@ use rustyline::error::ReadlineError;
 use rustyline::highlight::{Highlighter, MatchingBracketHighlighter};
 use rustyline::hint::{Hinter, HistoryHinter};
 use rustyline::validate::{self, MatchingBracketValidator, Validator};
-use rustyline::{Cmd, CompletionType, Config, Context, EditMode, Editor, KeyPress, Movement, Word};
+use rustyline::{Cmd, CompletionType, Config, Context, EditMode, Editor, KeyPress};
 use rustyline_derive::Helper;
 
 #[derive(Helper)]
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -6,7 +6,7 @@ use rustyline::error::ReadlineError;
 use rustyline::highlight::{Highlighter, MatchingBracketHighlighter};
 use rustyline::hint::{Hinter, HistoryHinter};
 use rustyline::validate::{self, MatchingBracketValidator, Validator};
-use rustyline::{Cmd, CompletionType, Config, Context, EditMode, Editor, KeyPress};
+use rustyline::{Cmd, CompletionType, Config, Context, EditMode, Editor, KeyPress, Movement, Word};
 use rustyline_derive::Helper;
 
 #[derive(Helper)]
@@ -83,7 +83,7 @@ fn main() -> rustyline::Result<()> {
     let config = Config::builder()
         .history_ignore_space(true)
         .completion_type(CompletionType::List)
-        .edit_mode(EditMode::Emacs)
+        .edit_mode(EditMode::Vi)
         .output_stream(OutputStreamType::Stdout)
         .build();
     let h = MyHelper {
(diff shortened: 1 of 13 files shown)


examples/example.rs
```diff
@@ -83,7 +83,7 @@ fn main() -> rustyline::Result<()> {
     let config = Config::builder()
         .history_ignore_space(true)
         .completion_type(CompletionType::List)
-        .edit_mode(EditMode::Emacs)
+        .edit_mode(EditMode::Vi)
```
gwenn, line 86: Please keep default `Emacs` mode.

examples/example.rs
```diff
@@ -97,6 +97,9 @@ fn main() -> rustyline::Result<()> {
     rl.set_helper(Some(h));
     rl.bind_sequence(KeyPress::Meta('N'), Cmd::HistorySearchForward);
     rl.bind_sequence(KeyPress::Meta('P'), Cmd::HistorySearchBackward);
+    rl.bind_sequence(KeyPress::ControlBackspace, Cmd::Kill(Movement::BackwardWord(1, Word::Vi)));
+    rl.bind_sequence(KeyPress::MetaBackspace, Cmd::Kill(Movement::BackwardChar(1)));
+    rl.bind_sequence(KeyPress::ShiftBackspace, Cmd::Kill(Movement::BackwardChar(1)));
```
gwenn, line 102: As these bindings don't work on unix, I would prefer that you remove them.

src/tty/windows.rs
```diff
@@ -179,6 +179,17 @@ impl RawReader for ConsoleRawReader {
...
+                        return Ok(if ctrl {
+                            KeyPress::ControlBackspace
+                        } else if shift {
+                            KeyPress::ShiftBackspace
+                        } else if meta {
+                            KeyPress::MetaBackspace
+                        } else {
+                            KeyPress::Backspace
+                        });
+                    }
```
gwenn, line 192: These lines are partially redundant with lines 235:236 and 247:248. So I would prefer you (re)move this block.

src/tty/windows.rs
```diff
@@ -220,13 +231,21 @@ impl RawReader for ConsoleRawReader {
                 };
                 let c = rc?;
                 if meta {
-                    return Ok(KeyPress::Meta(c));
+                    let mut key = keys::char_to_key_press(c);
```
gwenn, line 234: You can factorize duplicated lines 234 and 242 by moving it at line 233.

src/keys.rs
```diff
@@ -24,6 +24,8 @@ pub enum KeyPress {
     ControlRight,
     /// Ctrl-↑
     ControlUp,
+    /// Control Backspace
```
gwenn, line 27: Maybe we should warn that this key can/may not be used on unix.

examples/example.rs
fdncred: Understood.

examples/example.rs
fdncred: I left those in there so they could be tested with example.rs, but I'll remove them

src/tty/windows.rs
fdncred: understood.

src/tty/windows.rs
fdncred: understood.

src/keys.rs
fdncred: I wasn't sure if you wanted that comment on ControlBackspace only or ShiftBackspace and MetaBackspace, so I only put it where you called it out.

fdncred:
I believe I changed all your requested items. Please let me know if you have further feedback.

src/tty/windows.rs
```diff
@@ -219,14 +219,21 @@ impl RawReader for ConsoleRawReader {
...
+                    }
+                    return Ok(key);
                 } else {
-                    let mut key = keys::char_to_key_press(c);
                     if key == KeyPress::Tab && shift {
                         key = KeyPress::BackTab;
                     } else if key == KeyPress::Char(' ') && ctrl {
                         key = KeyPress::Ctrl(' ');
+                    } else if key == KeyPress::Backspace && ctrl {
+                        key = KeyPress::ControlBackspace;
```
gwenn, line 236: You forgot `ShiftBackspace` case.

src/lib.rs
```diff
@@ -888,7 +888,22 @@ impl<H: Helper> Editor<H> {
...
-            bindings.insert(key_seq, cmd)
+            let key = if let KeyPress::Char(ref c) = key_seq {
+                if c.is_control() {
+                    keys::char_to_key_press(*c)
+                } else {
+                    key_seq
+                }
+            } else if let KeyPress::Ctrl(ref c) = key_seq {
+                if c.is_control() {
+                    keys::char_to_key_press(*c);
```
gwenn, line 899: This line can be removed.

src/tty/windows.rs
fdncred: understood.

src/lib.rs
fdncred: understood.

src/tty/windows.rs
```diff
@@ -219,14 +219,23 @@ impl RawReader for ConsoleRawReader {
                     return Err(error::ReadlineError::Eof);
                 };
                 let c = rc?;
+                let mut key = keys::char_to_key_press(c);
                 if meta {
-                    return Ok(KeyPress::Meta(c));
+                    if key == KeyPress::Backspace && meta {
```
gwenn, line 224: `meta` is always `true`..

src/lib.rs
```diff
@@ -888,7 +888,21 @@ impl<H: Helper> Editor<H> {
...
-            bindings.insert(key_seq, cmd)
+            let key = if let KeyPress::Char(ref c) = key_seq {
+                if c.is_control() {
+                    keys::char_to_key_press(*c)
+                } else {
+                    key_seq
+                }
+            } else if let KeyPress::Ctrl(ref c) = key_seq {
+                if c.is_control() {
+                    warn!(target: "rustyline", "KeyPress::Ctrl({:?}) may not work on unix", c)
```
gwenn, line 899: In fact, based on current implementation, both unix and windows do not support such combination...

src/tty/windows.rs
fdncred: done

src/lib.rs
fdncred: I added these three items back in the example program and tested it and it performed as expected.
```rust
    rl.bind_sequence(KeyPress::ControlBackspace, Cmd::Kill(Movement::BackwardWord(1, Word::Big)));
    rl.bind_sequence(KeyPress::MetaBackspace, Cmd::Kill(Movement::BackwardWord(1, Word::Big)));
    rl.bind_sequence(KeyPress::ShiftBackspace, Cmd::Kill(Movement::BackwardWord(1, Word::Big)));
```
Please help me understand what you're talking about when you say, `both unix and windows do not support such a combination`

gwenn: Sorry, I mean `KeyPress::Crtl('\x08')`, for example, does not work on unix or windows...

fdncred: It's not supposed to work in this PR. That's why `ControlBackspace` was added.

gwenn: So the comment is misleading.

fdncred: I'm sorry @gwenn but I'm not following what you're saying. What comment is misleading? I was originally trying to add `KeyPress:Ctrl('\x08')` but you made it clear in this PR that it wouldn't work that way. So, I added `ControlBackspace`, `MetaBackspace`, `ShiftBackspace` in place of being able to take char literals.

gwenn: At line 899:
`KeyPress::Ctrl({:?}) may not work on unix`
should be rephrased:
`KeyPress::Ctrl({:?}) does not work`
or something like this.

fdncred: updated

fdncred:
@gwenn ping. are we ready to merge this?

gwenn:
I would prefer to fix and merge #333 before.

PR #424, opened 2020-08-02
Add History Search Behaviour

### LINKED ISSUE #423: Use Arrow-Up to search history with prefix
If I type something, and then press the arrow up key, I'd like that to search through the history, finding the next command that starts with the same prefix as there is before the cursor. For example if I type the following:
```
> cd /a/really/long/path
> ls
> cd /another/really/long/path
> ls
> cd /a/ ⬆
```
I'd like to get the first command autocompleted. Another example:
```
> cd abcdef
> ls
> cd ab123
> ls
> cd abc ⬅ ⬆
```
In that example, my cursor will be after `cd ab` before the last `c`. When pressing up, I'd now like to have suggested `cd ab123` and after another arrow-up-press I'd like to be suggested `cd abcdef`.

Is something like this already being worked on?

Reference: nushell/nushell#465

oberien:
Allow history search via Arrow-Up/Down or C-p/n to go either line-by-line or based on the prefix before the cursor position

Fix #423


src/edit.rs
```diff
@@ -614,13 +641,17 @@ impl<'out, 'prompt, H: Helper> State<'out, 'prompt, H> {
...
-            self.refresh_line()
+            let pos = if keep_cursor_pos {
+                self.line.pos()
+            } else {
+                buf.len()
+            };
+            self.update_line(buf, pos);
         } else {
-            self.out.beep()
+            self.restore();
```
oberien, line 651: I don't know if this `restore` breaks the following code:

https://github.com/kkawakam/rustyline/blob/88e669c4ff17c1e7cc59f5eeff12ac3dc068c5c6/src/lib.rs#L558-L559

oberien:
I rebased this PR to the newest master. Is there anything that's missing?

oberien:
I squashed my commits to make rebasing easier and rebased onto the current master.

The issue behind this PR in nushell comes up every now and again to allow nushell to have zsh-similar search behaviour taking into account the typed prefix. Is there anything I can do to help get this PR merged?

oberien:
@gwenn Is this PR something that could be of interest to rustyline?

fdncred:
@gwenn We would really like to have this on nushell. Can you please look at this PR?

src/edit.rs
```diff
@@ -625,13 +652,28 @@ impl<'out, 'prompt, H: Helper> State<'out, 'prompt, H> {
...
+            } else {
+                buf.len()
+            };
+            self.update_line(buf, pos);
+        } else {
+            self.ctx.history_index = match dir {
+                Direction::Forward => history.len(),
+                Direction::Reverse => 0,
+            };
+            self.restore();
```
gwenn, line 673: bash `history-search-backward` does not seem to loop or skip same history entries.
But you are right: cursor is preserved.

gwenn:
https://github.com/kkawakam/rustyline/pull/545/commits/eb68960f866f229c64f05aea5ca6570765a5daf9 preserves cursor position.

PR #430, opened 2020-08-14
Extend Meta-F,Alt+Right feature for hint partial completion

hiking90:
Inspired from Fish shell hint feature.


gwenn:
I know that currently you cannot perform conditional binding (#269).
But when #269 will be fixed, I would prefer to keep all these custom bindings either in your project or behind a `fish` feature.

hiking90:
Ok. I'll close this pull request.

gwenn:
Partially fixed in version 8.0.0
But you don't have access to hint content yet.
See https://github.com/kkawakam/rustyline/blob/master/examples/custom_key_bindings.rs#L53

PR #434, opened 2020-09-02
Fixed error detection in move_cursor_at_leftmost

### LINKED ISSUE #433: Windows still panics in some cases
Hi all,

Thanks for the fixes to, I think we managed to fix some of our Windows instability with this PR: https://github.com/kkawakam/rustyline/issues/359

Unfortunately, users are still reporting very similar issues. I haven't been able to reproduce it, yet, but I wanted to file an issue in case you are able to reproduce: https://github.com/nushell/nushell/issues/2440

Happy to help wherever we can.

Lucretiel:
Code 87 emits an `ErrorKind::InvalidInput`, not `ErrorKind::Other`. Fixed the detection here by removing the `ErrorKind` detection entirely and using a winapi named constant. Fixes #433. See https://github.com/rust-lang/rust/blob/master/library/std/src/sys/windows/mod.rs#L64 for details.

### DIFF
--- a/src/tty/windows.rs
+++ b/src/tty/windows.rs
@@ -9,6 +9,7 @@ use log::{debug, warn};
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
 use winapi::shared::minwindef::{BOOL, DWORD, FALSE, TRUE, WORD};
+use winapi::shared::winerror;
 use winapi::um::winnt::{CHAR, HANDLE};
 use winapi::um::{consoleapi, handleapi, processenv, winbase, wincon, winuser};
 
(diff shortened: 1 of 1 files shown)


Lucretiel:
Fixes #433

Lucretiel:
Yup, looks like ERROR_INVALID_PARAMETER was moved to InvalidInput a couple months ago https://github.com/rust-lang/rust/commit/b71a3e1e3a20e7db3b8167d19002c372ffe69c54

gwenn:
Thanks.
And do you know which cursor position is invalid ? X, Y or both ?

Lucretiel:
No idea; I'm not familiar enough with the relevant Windows API. I also don't know why in practice this error appears on Windows Terminal and Alacritty but not cmd.exe

jonathandturner:
Just confirming that this fixes the issue in Nushell. We'll keep testing, but aren't able to repro the failures we saw before.

@gwenn - would you be able to roll out a new rustyline release with this fix? That would let us move to it for the next Nushell release as well.

Lucretiel:
Unfortunately it seems like other people were having trouble even reproducing the issue in nushell, which surprised me. It's still under investigation.

jonathandturner:
@Lucretiel we managed to reproduce it this morning, luckily. Looks like this PR addresses the cases we could find.

gwenn:
Thanks you all for your help.
I will try to release a new version this week-end.

PR #448, opened 2020-10-04
Disable CPR by default

### LINKED ISSUE #445: Make possible to deactivate move_cursor_at_leftmost
See https://github.com/edgedb/edgedb-cli/issues/101
Also,  many issues are associated to this method (#433, #402, #359, ...).
And, as far as I know, similar librairies don't have such "feature" (except maybe liner).
=> Add a config flag

gwenn:
Fix #445

TODO:
- [ ] Validate deactivation by default

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -28,6 +28,8 @@ pub struct Config {
     output_stream: OutputStreamType,
     /// Horizontal space taken by a tab.
     tab_stop: usize,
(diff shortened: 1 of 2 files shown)


gwenn:
@jonathandturner, @tailhook Are you OK if we don't check that cursor is at left most position by default ?

PR #452, opened 2020-10-13
Implement indentation commands

tailhook:
They are currently only attached to vim commands `>` and `<` (as Tab is used for completion).

### DIFF
--- a/src/line_buffer.rs
+++ b/src/line_buffer.rs
@@ -1,6 +1,7 @@
 //! Line buffer with current cursor position
 use crate::keymap::{At, CharSearch, Movement, RepeatCount, Word};
 use std::cell::RefCell;
+use std::cmp::min;
 use std::fmt;
 use std::iter;
 use std::ops::{Deref, Index, Range};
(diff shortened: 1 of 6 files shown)


src/line_buffer.rs
```diff
@@ -1097,6 +1099,78 @@ impl LineBuffer {
         }
         killed
     }
+
+    /// Indent range specified by `mvt`.
+    pub fn indent(&mut self, mvt: &Movement, amount: usize, dedent: bool)
+        -> bool
+    {
+        let pair = match *mvt {
+            // All inline ooperators are the same: indent current line
```
gwenn, line 1108: typo `ooperators`

gwenn:
`python-prompt-toolkit` and `jline3` seem to provide this feature.
A conditional binding is needed to make `Tab` key behave correctly.

src/line_buffer.rs
tailhook: Fixed

tailhook:
Fixed a typo in comment. I think conditional Tab can be added in subsequent PRs.

Issue #453, opened 2020-10-13
How to implement auto-indentation?


tailhook:
Hi!

I want to implement auto-indentation for [EdgeDB REPL](https://github.com/edgedb/edgedb-cli/). That is if you have:
```
edgedb> SELECT User {▮<enter>
```
it get's next line already indented:
```
edgedb> SELECT User {
.......     ▮
```
(where ▮ denotes cursor)

My idea was to implement indentation commands (#452) and then implement the [`invoke` method of the `ValidatorContext`](https://github.com/kkawakam/rustyline/blob/5596326c49098f47da371de60ceff7b64fbdefcc/src/validate.rs#L33-L36). The problem is that it requires refactoring the whole input loop around: https://github.com/kkawakam/rustyline/blob/5596326c49098f47da371de60ceff7b64fbdefcc/src/lib.rs#L497-L705

Is this what is need to be done? (I'm asking because it's quite complex PR). Or is there any better ideas of how to do auto-indentation?

gwenn:
We already need to fix the main loop for composite command (see https://github.com/kkawakam/rustyline/pull/293#issuecomment-557862724).
But I am not sure how to support auto-indentation, reformat, ...

PR #455, opened 2020-10-16
Factored out execute function for the main loop

tailhook:
Most commands are factored out to the function, except:

1. Commands that need extra input
2. Suspend, which needs access to the terminal

This should be enough to implement `ContextValidator::invoke`.

Also this commit brings `CompleteHint`, `SelfInsert`, `Insert` commands
to the `match` over `Cmd` instead of keeping them as separate `if`
statements as latter doesn't seem to change their behavior.

This is needed to implement #453, and #293 (although, latter might need
input-accepting commands too or another approach).

-----

This is somewhat super-minimal refactoring to move closer to our goal of invoking commands from `ContextValidator::invoke` or any other user hooks.

I think `input_mode.is_emacs_mode()` and a reference to `kill_ring` can also be incorporated into `State` object. So the only thing we need to propagate through the call chain is `State`. What do you think? (this can be a follow-up PR anyway)

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -1,0 +1,235 @@
+use std::sync::{Arc, Mutex};
+
+use crate::{Helper, Result};
+use crate::{complete_hint_line};
+use crate::config::Config;
+use crate::edit::State;
+use crate::error;
+use crate::history::{Direction};
+use crate::keymap::{Anchor, At, Cmd, Movement, Word};
+use crate::kill_ring::{KillRing, Mode};
+use crate::line_buffer::WordAction;
+use crate::keymap::{InputState, Refresher};
+
+
+pub enum Status {
+    Proceed,
+    Submit,
+}
+
+pub fn execute<H: Helper>(
+    cmd: Cmd,
+    s: &mut State<'_, '_, H>,
+    input_state: &InputState,
+    kill_ring: &Arc<Mutex<KillRing>>,
+    config: &Config,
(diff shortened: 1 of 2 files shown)


gwenn:
For #293 (more precisely, for composite command), we need to cumulate impacts of each action and update output accordingly only once.
But your PR is still useful.

tailhook:
> For #293 (more precisely, for composite command), we need to cumulate impacts of each action and update output accordingly only once.

Well, I think we can do the following:

1. Drop most `move_cursor` and `refresh_*` in the `edit.rs`
2. Refresh line by looking at growing `changes`
3. If line is not refreshed move cursor by comparing to old position

(sans few special commands, like `Undo`)
Then `ValidateContext::invoke` also would only do the "impact" and and line is refreshed afterwards.

Thoughts?

> But your PR is still useful.

Okay. Let's merge it and proceed quickly. Because this is a kind of change which will conflict with every addition or even change or other command (including #452) and keeping track of which commands are added in the master and which are already there in this branch is time-consuming.

tailhook:
Rebased on the new master. So do you accept this as a first step? As I've said rebasing this PR every time a big piece of error-prone work so I don't want to keep it in a branch for too long.

gwenn:
Sorry, the conflict was introduced by your PR #452...

gwenn:
Thanks.

PR #458, opened 2020-10-20
Ignore EINTR while polling

gwenn:


src/tty/unix.rs
```diff
@@ -608,7 +608,18 @@ impl PosixRawReader {
...
         let mut fds = [poll::PollFd::new(STDIN_FILENO, PollFlags::POLLIN)];
-        poll::poll(&mut fds, timeout_ms)
+        let r = poll::poll(&mut fds, timeout_ms);
+        match r {
+            Ok(_) => r,
+            Err(nix::Error::Sys(nix::errno::Errno::EINTR)) => {
+                if SIGWINCH.load(atomic::Ordering::Relaxed) {
+                    r
+                } else {
+                    Ok(0) // Ignore EINTR while polling
```
bjorn3, line 618: This doesn't offer any way to get the old behavior back. The old behavior os useful for shells where ctrl-c would result in EINTR. It should cancel the input in that case.

gwenn: Are you sure that Ctrl-C still generates an interrupt when `BRKINT` is disabled ?

bjorn3: `stty sane` enables `BRKINT`. Also with a shell I am not only referring to something like bash, but also repls like `python`, `node`, `gdb`, ...

gwenn: `linefeed`/`mortal` catches EINTR (but only on SIGWINCH):
https://github.com/murarth/mortal/blob/master/src/unix/terminal.rs#L477-L486
```rust
                Err(ref e) if e.as_errno() == Some(Errno::EINTR) => {
                    if get_signal().is_some() {
                        return Ok(true);
                    }
                }
```
`replxx` catches EINTR (but continues):
https://github.com/AmokHuginnsson/replxx/blob/master/src/io.cxx#L277-L279
```c++
		do {
			nread = read( STDIN_FILENO, &c, 1 );
		} while ((nread == -1) && (errno == EINTR));
```
https://github.com/AmokHuginnsson/replxx/blob/master/src/io.cxx#L594-L597
```c++
		int err( select( nfds, &fdSet, nullptr, nullptr, timeout_ > 0 ? &tv : nullptr ) );
		if ( ( err == -1 ) && ( errno == EINTR ) ) {
			continue;
		}
```

`rustyline` already catches EINTR except on SIGWINCH but only while reading (not polling):
https://github.com/kkawakam/rustyline/blob/master/src/tty/unix.rs#L134-L138
```rs
                if error.kind() != io::ErrorKind::Interrupted
                    || SIGWINCH.load(atomic::Ordering::Relaxed)
                {
                    return Err(error);
                }
```
So this PR only applies the same logic while polling !

gwenn: @bjorn3 Could you please provide a use case where this PR generates a regression ?
Thanks.

bjorn3: Never mind, I hadn't seen https://github.com/kkawakam/rustyline/blob/e9a5ccdc4f24c4a73053e6cf8897c186da5981da/src/keymap.rs#L959

PR #460, opened 2020-11-15
Document Validator::validate_while_typing not being implemented

lberrymage:
This was previously not apparent to library consumers, and can/has lead
to confusion when the function doesn't/didn't work as users expect(ed).
This commit makes it more clear to users that the function is a no-op by
exposing such information through rustdoc.


gwenn:
Thanks.

PR #407, opened 2020-06-20
Use Ctrl-E for hint completion

AndreaOrru:
See #406

Issue #406, opened 2020-06-20
Complete based on hints


AndreaOrru:
Assuming `Hinter` is designed to work similarly to the completion mechanism offered by `fish` and `prompt-toolkit`, I believe it should offer a simple way to complete to the suggested hint.

In `fish`, you can press `Ctrl-E`, and it will bring you to the end of the line _including the hint_. That doesn't work with `rustyline`. Am I missing something?

gwenn:
See `CompleteHint` command:
https://docs.rs/rustyline/6.2.0/rustyline/enum.Cmd.html#variant.CompleteHint
Mapped to `Right` key by default:
https://github.com/kkawakam/rustyline/blob/be20cfac991e7bc30d59b712b219438f5222f6e5/src/keymap.rs#L496

You can associate `Ctrl-E` to `CompleteHint` with:
```rust
    let mut rl = Editor::with_config(config);
    rl.bind_sequence(KeyPress::Ctrl('E'), Cmd:: CompleteHint);
```

AndreaOrru:
Hi @gwenn, thanks for your answer. That's not what I want. `Ctrl-E` may also be used to go to the end of the line when no hint is available - the binding you're suggesting overrides that behaviour.

gwenn:
See [highlight_hint](https://docs.rs/rustyline/6.2.0/rustyline/highlight/trait.Highlighter.html#method.highlight_hint)
and [highlight](https://docs.rs/rustyline/6.2.0/rustyline/highlight/trait.Highlighter.html#method.highlight) for 2 & 3.
And see #306 (and associated PR) to bind one action conditionally.

AndreaOrru:
Thank you! I solved my issues 2 & 3 with `highlight_hint`.

I appreciate the PR you linked would solve my issue n. 1. However, I still think `Ctrl-E` should complete the hint by default, like in `fish` and `prompt-toolkit`.

Issue #306, opened 2019-12-08
Multiple commands for a keybinding


adam-mcdaniel:
It would be a really handy feature to add a `Cmd::Multiple(Vec<Cmd>)` enumeration for the `Cmd` enum so that multiple commands could be used in succession for a keybinding.

gwenn:
WIP: #293

nu8:
https://github.com/nushell/nushell/issues/2421

89z:
https://github.com/nushell/nushell/issues/2781

89z:
https://github.com/kkawakam/rustyline/pull/492

89z:
so should this be closed since #492 was merged?

gwenn:
> so should this be closed since #492 was merged?

No, #492 makes possible to put condition(s) on which command is executed depending on input state.
But still only one command can be executed.

superiums:
still looking for this feature. a common use case is: after replace string, we need move cursor.  is there any way to finish this ?

hdwalters:
> I see that enum EventHandler has this TODO comment

With that in mind, I would not recommend adding a new `EventHandler::Macro` enum value; it seems much better to add a new `Cmd::Multiple(Vec<Cmd>)` as per the OP's request. Note, it would have to be `Cmd::Multiple(Vec<Box<Cmd>>)` because `Cmd` cannot be sized with this addition.

PR #466, opened 2020-12-03
Initial invoke trait and auto-indent example

tailhook:
Here is a somewhat minimal implementation of invoking commands from validator trait.

General notes: it's unclear what to do with `command::Status` which can be `Proceed` or `Submit`. Currently it's ignored. But we may want to disable (i.e. assert on) commands returning `Submit` in validator and don't expose `Status` type to the user?

About auto-indent (#453), I've put mostly working example, but there are few ugly parts of the implementation:
1. Data validated without newline. So to indent the new line, validator inserts line manually
2. Which is a problem as any NewLine command triggers validation again (infinite recursion), so I protect for that in validator itself using RefCell, but this seems ugly. Should we disable validation for commands run from validator? Or in future anywhere except typed by keys?
3. After inserting NewLine i need avoid it inserting by the execute trait. I use "empty message trick". But I'm not sure it's future-proof.

So while auto-indentation is just one use case for `invoke`, I think at least (3) should be fixed somehow. Quick options are:
1. Insert newline before validation (probably breaks too much stuff)
2. Add a validator option method, like `validate_after_newline() -> bool { false }` that makes it optional for validator. Or (2a) have a config option, or (2b) have different method for post-validation.
3. Add some `ValidationContext::skip_newline()` method, to skip adding newline after validation

(note: (1) and (2) fix `RefCell` issue too, unlike (3))

### DIFF
--- a/examples/auto_indent.rs
+++ b/examples/auto_indent.rs
@@ -1,0 +1,69 @@
+use std::cmp::Ordering;
+use std::io::Write;
+use std::cell::RefCell;
+
+use rustyline::error::ReadlineError;
+use rustyline::validate::{ValidationContext, ValidationResult, Validator};
+use rustyline::Editor;
+use rustyline_derive::{Completer, Helper, Highlighter, Hinter};
+use rustyline::{Cmd, Movement};
+
+#[derive(Completer, Helper, Highlighter, Hinter)]
+struct InputValidator {
+    stack_protect: RefCell<()>,
+}
+
+const INDENT_SIZE: usize = 2;  // default
+
+impl Validator for InputValidator {
+    fn validate(&self, ctx: &mut ValidationContext) -> Result<ValidationResult, ReadlineError> {
+        let input = ctx.input();
+        let _protect = if let Ok(cell) = self.stack_protect.try_borrow_mut() {
+            cell
+        } else {
+            // This is reached when `NewLine` is called below
+            return Ok(ValidationResult::Valid(None));
+        };
+        if let Some(line) = input.lines().last() {
+            let indent_chars = line.len() - line.trim_start().len();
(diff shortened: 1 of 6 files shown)


gwenn:
Related issues:
#269
Maybe #306 ?
#406
Related PRs:
#293
#342
#430
#407

gwenn:
> Which is a problem as any NewLine command triggers validation again

https://github.com/kkawakam/rustyline/pull/293#issuecomment-557862724
Currently, we have: Key -> Command -> Logic
If we do: Key -> Command -> Action(s) -> Logic and make sure `invoke` can only call Actions,
Then we should be fine.
However, this is a huge refactoring...

["All problems in computer science can be solved by another level of indirection" / "But it also can add complexity to the system, and therefore is not a free lunch."](https://wiki.c2.com/?OneMoreLevelOfIndirection)

tailhook:
I'm not sure i understand. Do you think that validation works on "Command" but invoke triggers "Action(s)"? This may be a problem too. What if actions changed text in a way that change the result of a validation?

Anyway, what next step(s) you anticipate? Do we make all this layer of indirection all at once?

gwenn:
> I'm not sure i understand. Do you think that validation works on "Command" but invoke triggers "Action(s)"? This may be a problem too. What if actions changed text in a way that change the result of a validation?

Even if you custom mapping/command triggers many actions, refresh/repaint and validation should be performed only once depending on the final result.

> Anyway, what next step(s) you anticipate? Do we make all this layer of indirection all at once?

I am afraid we have no choice.
And if possible, you should also try to support all kind of [filters](https://python-prompt-toolkit.readthedocs.io/en/master/pages/advanced_topics/key_bindings.html?highlight=filter#attaching-a-filter-condition).

PR #468, opened 2020-12-04
Migrated to dirs-next due to dirs no longer being maintained

### LINKED ISSUE #467: `dirs` is no longer maintained
Hello,
The `dirs` crate is no longer maintained, and the projeect migrated to `dirs-next` in #388. There is a comment there that seems to imply it's active, but as you can see in [its repository](https://github.com/dirs-dev/dirs-rs), it's been archived. Also, the newest versions of `dirs-next` are newer to those in `dirs`, so this part of #456 should probably be reverted.

Note that there is a RUSTSEC advisory about it: https://rustsec.org/advisories/RUSTSEC-2020-0053.html

Razican:
This fixes #467.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ members = ["rustyline-derive"]
 [dependencies]
 bitflags = "1.2"
 cfg-if = "1.0"
-dirs = { version = "3.0", optional = true }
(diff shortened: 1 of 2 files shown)


gwenn:
Thanks.

Issue #475, opened 2020-12-14
German keymap does not work correctly on Windows 10


tofpie:
In the German keymap some characters (like |) are inserted by pressing the modifier key `Alt Gr`. This does not work in rustyline.

It seems that `Alt Gr` (which is `Alt` + `Ctrl`) is interpreted as `Ctrl` on Windows:

https://github.com/kkawakam/rustyline/blob/4a87a1944ee73bcad760b44699e145ee64657a83/src/tty/windows.rs#L140-L152

gwenn:
7.1.0 released.

PR #476, opened 2020-12-14
Fix issue: AltGr handled as Ctrl

tofpie:
This fixes issue #475 about `AltGr` handling.


gwenn:
Thanks.
LGTM
I will check.
For reference:  https://github.com/AmokHuginnsson/replxx/blob/master/src/terminal.cxx#L374-L387

gwenn:
Validated

PR #486, opened 2021-02-16
Add ClearBuffer command

sigmaSd:
Hello,

This command is useful for repl(s).

For example IPython ctrlc does this, (it doesn't exit the shell, it only clears the buffer)


sigmaSd:
I found out that the default behavior of Interrupt is similar(but not quite the same), so I guess no need for this for now.

PR #505, opened 2021-03-23
Remove dependency on unmaintained fs2

coolreader18:
fs2 hasn't made a release in 3 years, and it looks like we only need it for file locking, which is pretty easy to do ourselves. Plus, this fixes compilation on Redox (danburkert/fs2-rs#38)

PR #507, opened 2021-03-28
Replace fs2 dependency by fd-lock

gwenn:


gwenn:
Should fix #505

coolreader18:
Yep, this looks good 👍

PR #435, opened 2020-09-06
Split KeyPress into separate Key and Modifier types.

gwenn:
Validate:
* Pros and cons of tuple struct versus  struct ?
* Tuple struct order: `(key, mods)` versus `(mods, key)` ?
* Remove `BackTab` and keep only `SHIFT-Tab` ?
* Remove `Modifiers::CTRL_SHIFT`, ... ?
* Normalize:
    - Ctrl-a => Ctrl-A
    - Shift-A => A

Todo:
\E\E[ sequences on unix (we should use `poll`).

This PR replaces/enhances:
* #333 (except \E\E[ sequences)
* #422
* #121 (should we support `SUPER` modifier ?)

This PR should fixes:
* #318
* #421

PR #517, opened 2021-05-03
Support ESC ESC sequences, and bind alt+left/alt+right

thomcc:
This has a few more of the sequences from #333. (Note that while #435 indicated this was the only type of missing sequence, I think there are a few more still).

Anyway, in particular, this style of sequence is emitted by iTerm2 (at least, in, err, the configuration I use, which I *think* is default), and so not having support for it was a little annoying. It's also used by rxvt-family terminals, and while I've tested this in them, I don't regularly use them, so I don't care as much.

Using recursion the way this does is a bit hacky way, but in practice it works on everything I've tried. Also, this follows the actual "logic" these terminals use for the meta/alt key, so it makes a kind of sense.

---

The other thing it implements is ALT+Left and ALT+Right by-word movement. This is pretty ubiquitous (emacs, bash, every macos textbox, ...), and was the main reason I wanted to make ESC ESC sequences work, so I added it too.

### DIFF
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -617,7 +617,10 @@ impl InputState {
             }
             E(K::Char('<'), M::ALT) => Cmd::BeginningOfHistory,
             E(K::Char('>'), M::ALT) => Cmd::EndOfHistory,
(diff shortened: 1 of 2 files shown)


gwenn:
Thanks.

gwenn:
Version 8.1.0 released

PR #518, opened 2021-05-05
Support multiline inputs in non TTY contexts

basile-henry:
This PR add supports for multiline inputs (via the `Validator` API) for the case where stdin isn't a TTY (or is an unsupported terminal).

I have added a new example `input_multiline` to make quickly showcase how this works.

This lets us do:
```
echo -e "(\n\n)" | cargo run --example input_multiline
```

And have readline correctly read this as a single multiline input instead of 3 separate inputs (if it was called in a loop).
Even in the single line case, there was a difference between the TTY and the non-TTY versions of readline: the non-TTY version would keep the trailing newline character.

My main motivation for this change is to be able to easily test a REPL made using `rustyline` by piping my inputs in without having to mock the TTY part of it. It also make `rustyline` more consistent this way.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -618,10 +618,24 @@ fn readline_direct(
         match reader.read_line(&mut input)? {
             0 => return Err(error::ReadlineError::Eof),
             _ => {
+                // Remove trailing newline
+                let trailing_n = input.ends_with('\n');
+                let trailing_r;
+
+                if trailing_n {
+                    input.pop();
+                    trailing_r = input.ends_with('\r');
+                    if trailing_r {
+                        input.pop();
+                    }
+                } else {
+                    trailing_r = false;
+                }
+
                 input = apply_backspace_direct(&input);
 
                 match validator.as_ref() {
-                    None => break,
+                    None => return Ok(input),
                     Some(v) => {
                         struct SimpleInvoke<'a> {
                             s: &'a str,
@@ -633,13 +647,7 @@ fn readline_direct(
                             }
                         }
 
-                        let mut ctx = SimpleInvoke {
-                            // Strip trailing newline for validation
-                            s: input
-                                .strip_suffix('\n')
-                                .map(|s| s.strip_suffix('\r').unwrap_or(s))
-                                .unwrap_or(&input),
-                        };
+                        let mut ctx = SimpleInvoke { s: &input };
                         let mut ctx = validate::ValidationContext::new(&mut ctx);
 
                         match v.validate(&mut ctx)? {
@@ -647,10 +655,19 @@ fn readline_direct(
                                 if let Some(msg) = msg {
                                     writer.write_all(msg.as_bytes())?;
                                 }
-                                break;
+                                return Ok(input);
                             }
                             validate::ValidationResult::Invalid(Some(msg)) => {
                                 writer.write_all(msg.as_bytes())?;
+                            }
+                            validate::ValidationResult::Incomplete => {
+                                // Add newline and keep on taking input
+                                if trailing_r {
+                                    input.push('\r');
+                                }
+                                if trailing_n {
+                                    input.push('\n');
+                                }
                             }
                             _ => {}
                         }
@@ -659,16 +676,6 @@ fn readline_direct(
             }
         }
     }
-
-    // Remove trailing newline
-    if input.ends_with('\n') {
-        input.pop();
-        if input.ends_with('\r') {
-            input.pop();
-        }
-    }
-
-    Ok(input)
 }
 
 /// Syntax specific helper.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -585,16 +585,8 @@ fn readline_direct(validator: &Option<impl Validator>) -> Result<String> {
         match io::stdin().read_line(&mut input)? {
             0 => return Err(error::ReadlineError::Eof),
             _ => {
-                // Remove trailing newline
-                if input.ends_with('\n') {
-                    input.pop();
-                    if input.ends_with('\r') {
-                        input.pop();
-                    }
-                }
-
                 match validator.as_ref() {
-                    None => return Ok(input),
+                    None => break,
                     Some(v) => {
                         struct SimpleInvoke<'a> {
                             s: &'a str,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -578,12 +578,55 @@ fn readline_raw<H: Helper>(
...
+                                // Add a newline back onto the end of the buffer
+                                input.push('\n');
+                            }
+                            validate::ValidationResult::Invalid(_msg) => {
+                                // TODO: present the msg back to stdout?
+                                return Ok(input);
+                            }
+                            validate::ValidationResult::Valid(_msg) => {
+                                // TODO: present the msg back to stdout?
+                                return Ok(input);
```
basile-henry, line 623: I am not entirely sure how to handle the case where `Valid` or `Invalid` return a message to the user.

Also `Invalid` maybe shouldn't return the input just yet :thinking:

gwenn: We may find an answer on how to handle message in [python-prompt-toolkit](https://python-prompt-toolkit.readthedocs.io/en/master/pages/asking_for_input.html#input-validation).

basile-henry: It looks like it ignores the message as well when not in TTY mode:

```bash
➜ echo -e 'hello\b\b\b\b\b45' | python test.py
Warning: Input is not a terminal (fd=0).
^[[65;1RGive a number: 45
You said: 45
```

basile-henry: Nevermind, I forgot that I need to "press Enter" before erasing the string "hello". It crashes, if I try that :stuck_out_tongue_closed_eyes:

basile-henry:
```bash
➜ echo -e 'hello\n\b\b\b\b\b45' | python test.py
Warning: Input is not a terminal (fd=0).
^[[65;1RGive a number: 45hello
Traceback (most recent call last):
  File "test.py", line 20, in <module>
    number = int(prompt('Give a number: ', validator=NumberValidator()))
  File "/home/basile/dev/temp/venv/lib/python3.8/site-packages/prompt_toolkit/shortcuts/prompt.py", line 1397, in prompt
    return session.prompt(
  File "/home/basile/dev/temp/venv/lib/python3.8/site-packages/prompt_toolkit/shortcuts/prompt.py", line 1020, in prompt
    return self.app.run(
  File "/home/basile/dev/temp/venv/lib/python3.8/site-packages/prompt_toolkit/application/application.py", line 887, in run
    return loop.run_until_complete(
  File "/usr/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/home/basile/dev/temp/venv/lib/python3.8/site-packages/prompt_toolkit/application/application.py", line 816, in run_async
    return await _run_async2()
  File "/home/basile/dev/temp/venv/lib/python3.8/site-packages/prompt_toolkit/application/application.py", line 792, in _run_async2
    result = await _run_async()
  File "/home/basile/dev/temp/venv/lib/python3.8/site-packages/prompt_toolkit/application/application.py", line 733, in _run_async
    result = await f
EOFError
```

basile-henry:
I have also added support to handle backspace characters in non TTY mode.

This makes the following work:

```bash
➜ echo -e "([)\n\b\b\n])" | cargo run --example input_multiline 2> /dev/null
Input: ([
])
```

src/lib.rs
basile-henry: Ok, I believe this is now properly handled.

src/lib.rs
```diff
@@ -692,13 +776,13 @@ impl<H: Helper> Editor<H> {
...
 
-            readline_direct()
+            readline_direct(io::stdin().lock(), io::stdout(), &self.helper)
         } else if self.term.is_stdin_tty() {
             readline_raw(prompt, initial, self)
         } else {
             debug!(target: "rustyline", "stdin is not a tty");
             // Not a tty: read from file / pipe.
-            readline_direct()
+            readline_direct(io::stdin().lock(), io::stdout(), &self.helper)
```
basile-henry, line 785: `io::stderr()` might be more appropriate for the validator messages :thinking:

basile-henry:
> there should be no ANSI escape sequence (colors) in the input stream

I don't really understand what you mean. I don't think I have added anything ANSI dependent in the input stream. Are you referring to `'\b'`? I believe that is a ASCII control character. And by the looks of it `python-prompt-toolkit` also supports it.

Or are you referring to the output stream, and I need to filter out ANSI sequences in the messages from the validator?

gwenn:
> > there should be no ANSI escape sequence (colors) in the input stream
> 
> I don't really understand what you mean. I don't think I have added anything ANSI dependent in the input stream. Are you referring to `'\b'`? I believe that is a ASCII control character. And by the looks of it `python-prompt-toolkit` also supports it.
> 
> Or are you referring to the output stream, and I need to filter out ANSI sequences in the messages from the validator?

Sorry I was completely wrong !

src/lib.rs
```diff
@@ -578,13 +578,97 @@ fn readline_raw<H: Helper>(
...
+                                self.s
+                            }
+                        }
+
+                        let mut ctx = SimpleInvoke {
+                            // Strip trailing newline for validation
+                            s: input
+                                .strip_suffix('\n')
+                                .map(|s| s.strip_suffix('\r').unwrap_or(s))
+                                .unwrap_or(&input),
```
gwenn, line 641: Are you sure ? `Validator` should support '\n' in input.

basile-henry: The validator does support newlines. The issue here is that in interactive mode the validator is run without the trailing newline, so I just want this to be consistent.
I am guessing in interactive mode the newline is only added after the validator is run.
The newline here is not removed from the buffer, and it will end up in the validator input if the user is entering multiple lines.

Maybe the newline should be removed if the validator returns `Invalid` :thinking: 
In an earlier commit I was removing the trailing newline in the buffer all the time and adding it back on in the `Incomplete` case. The reason I changed it is that I was dropping `\r` while doing so.
`Invalid` is an awkward case to handle in non TTY mode anyway. :sweat_smile:

basile-henry: Ok, I believe I have fixed that edge case. I think this issue is a matter of thinking of `\n` as a control character in the input sequence and not as a character that happens to be part of the input.

basile-henry:
Since I have fixed an edge case in the way the `Invalid` case worked the quick example is now this:

```bash
➜ echo -e "([)\n\b\n])" | cargo run --example input_multiline 2> /dev/null
Input: ([
])
```

Notice how I don't need to backspace the `\n` anymore, because the input is invalid the newline isn't entered, so `\b` char applies to `)` instead.

src/lib.rs
```diff
@@ -578,12 +578,103 @@ fn readline_raw<H: Helper>(
...
+                    trailing_r = false;
+                }
+
+                input = apply_backspace_direct(&input);
+
+                match validator.as_ref() {
+                    None => return Ok(input),
+                    Some(v) => {
+                        struct SimpleInvoke<'a> {
+                            s: &'a str,
```
gwenn, line 641: See https://github.com/kkawakam/rustyline/pull/534/files

gwenn:
Thanks and sorry for the long delay.

basile-henry:
No worries, thanks for reviewing and taking the feature in! :smile:

gwenn:
Version 9.0.0 released.

Issue #520, opened 2021-05-08
fd-lock change breaks build on windows


marcbowes:
56b925c271b56da77e7a78a64470e239bc1daa09 breaks the build on my Windows machine.

```
   Compiling dirs-next v2.0.0
error[E0308]: mismatched types
  --> C:\Users\...\.cargo\registry\src\github.com-1ecc6299db9ec823\fd-lock-2.0.0\src\windows.rs:39:33
   |
39 |         if unsafe { !UnlockFile(handle, 0, 0, 1, 0) } == 0 {
   |                                 ^^^^^^ expected enum `winapi::ctypes::c_void`, found enum `std::ffi::c_void`
   |
   = note: expected raw pointer `*mut winapi::ctypes::c_void`
              found raw pointer `*mut std::ffi::c_void`

error[E0308]: mismatched types
  --> C:\Users\...\.cargo\registry\src\github.com-1ecc6299db9ec823\fd-lock-2.0.0\src\windows.rs:73:35
   |
73 |         match unsafe { LockFileEx(handle, flags, 0, 1, 0, overlapped.raw()) } {
   |                                   ^^^^^^ expected enum `winapi::ctypes::c_void`, found enum `std::ffi::c_void`
   |
   = note: expected raw pointer `*mut winapi::ctypes::c_void`
              found raw pointer `*mut std::ffi::c_void`

error[E0308]: mismatched types
  --> C:\Users\...\.cargo\registry\src\github.com-1ecc6299db9ec823\fd-lock-2.0.0\src\windows.rs:90:33
   |
90 |         match unsafe { LockFile(handle, 0, 0, 1, 0) } {
   |                                 ^^^^^^ expected enum `winapi::ctypes::c_void`, found enum `std::ffi::c_void`
   |
   = note: expected raw pointer `*mut winapi::ctypes::c_void`
              found raw pointer `*mut std::ffi::c_void`
```

marcbowes:
It seems the commit can be cleanly reverted, after which the build succeeds again.

gwenn:
CI build seems OK on windows: https://github.com/kkawakam/rustyline/runs/2486841046?check_suite_focus=true

gwenn:
https://docs.rs/fd-lock/2.0.0/x86_64-pc-windows-msvc/fd_lock/index.html seems OK too.

marcbowes:
Pretty weird!

gwenn:
[...]

Issue #519, opened 2021-05-07
How to make CTRL-V paste on windows (cmd.exe, powershell)


marcbowes:
When using rustyline on Linux or OSX, system copy-paste works just fine. However, on Windows that's not the case.

Here's a scenario that doesn't work as expected:

1. Open notepad and type some text. Select it, use CTRL-C to put it on the clipboard.
2. Open cmd.exe and press CTRL-V. The text should paste.
3. Open a rustyline application (e.g. the default example will do). Press CTRL-V. The next *will not paste*.
4. Use rightclick. The text will paste.

The same thing is true when using Powershell. When using Windows Terminal, CTRL-V works as expected.

I looked through the available keybinds/commands and I don't see a "paste from system clipboard" anywhere. Should this work by default, or is there something I need to enable?

gwenn:
I only have access to a Windows machine at work.
To check:
- how does [linefeed](https://github.com/murarth/linefeed) / [termwiz](https://github.com/wez/wezterm/tree/main/termwiz) / [replxx](https://github.com/AmokHuginnsson/replxx) behave ?
- For rustyline, by default, CTRL-V does nothing on Windows and there is no command to "paste from system clipboard" so you cannot use [bind_sequence](https://docs.rs/rustyline/8.0.0/rustyline/struct.Editor.html#method.bind_sequence).
- Maybe you should try to tweak Command / Powershell settings.

marcbowes:
Well, that's pretty clear! Thanks.

I know literally nothing about the windows API. If you can point me at some method names, I can try implement it.

marcbowes:
https://github.com/marcbowes/rustyline/commit/b0e2669acacf9c259f0f73bc1d9d8f271f8a07c2

I'll test that out later on my Windows machine. If it works, I'll wire up a paste command.

marcbowes:
See the above PR for a fix. Not sure if I did it right. Also had to revert the fd-lock change to get things compiling.

gwenn:
Version 8.1.0 released

marcbowes:
Thank you!

PR #521, opened 2021-05-08
Implement CTRL-V support on Windows

marcbowes:
This commit implements CTRL-V support on Windows. See #519 where this issue was raised. 

## Implementation notes

On Windows, we have 3 key changes:

* There are new Windows-only dependencies on the `clipboard-win` and `error-code` crates.
* tty/windows.rs implements `read_pasted_text` by calling the clipboard-win crate. Errors are mapped to a new Windows-only error variant.
* When CTRL-V is pressed, `QuotedInsert` will perform a yank using the text from the clipboard (inserting the text and moving the cursor).

## Testing

I made a new package:

```
cargo new testpaste
cd testpaste
cargo add rustyline
# edit Cargo.toml; make rustyline use my git fork
# edit src/lib.rs to be the contents of https://github.com/kkawakam/rustyline#example
```

As noted in #519, running this example in Windows Terminal works (CTRL-V pastes text from the system clipboard) while using Powershell or cmd.exe does not work (CTRL-V does nothing, while rightclick pastes text from the system clipboard).

After this change, CTRL-V works in all 3 shells. Mouse rightclick continues to work.

## Other

See #520 for why I reverted the change to use fd-lock.

### DIFF
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -28,6 +28,9 @@ pub enum Cmd {
     CapitalizeWord,
     /// clear-screen
     ClearScreen,
+    /// Paste from the clipboard
+    #[cfg(windows)]
+    PasteFromClipboard,
     /// complete
     Complete,
     /// complete-backward
@@ -1046,9 +1049,13 @@ impl InputState {
                 } else {
                     Cmd::Kill(Movement::EndOfLine)
                 }
-            },
-            E(K::Char('Q'), M::CTRL) | // most terminals override Ctrl+Q to resume execution
+            }
+            // most terminals override Ctrl+Q to resume execution
+            E(K::Char('Q'), M::CTRL) => Cmd::QuotedInsert,
+            #[cfg(not(windows))]
             E(K::Char('V'), M::CTRL) => Cmd::QuotedInsert,
+            #[cfg(windows)]
+            E(K::Char('V'), M::CTRL) => Cmd::PasteFromClipboard,
             E(K::Char('W'), M::CTRL) => {
                 if positive {
                     Cmd::Kill(Movement::BackwardWord(n, Word::Big))
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -1034,10 +1034,9 @@ impl InputState {
                     Cmd::Move(Movement::BackwardChar(n))
                 }
             }
-            E(K::Char('J'), M::CTRL) |
-            E::ENTER => {
-                Cmd::AcceptOrInsertLine { accept_in_the_middle: true }
-            }
+            E(K::Char('J'), M::CTRL) | E::ENTER => Cmd::AcceptOrInsertLine {
+                accept_in_the_middle: true,
+            },
             E(K::Down, M::NONE) => Cmd::LineDownOrNextHistory(1),
             E(K::Up, M::NONE) => Cmd::LineUpOrPreviousHistory(1),
             E(K::Char('R'), M::CTRL) => Cmd::ReverseSearchHistory,
@@ -1076,10 +1075,10 @@ impl InputState {
             E(K::BracketedPasteStart, M::NONE) => {
                 let paste = rdr.read_pasted_text()?;
                 Cmd::Insert(1, paste)
-            },
-            _ => {
-                self.custom_seq_binding(rdr, wrt, &mut evt, n, positive)?.unwrap_or(Cmd::Unknown)
-            },
+            }
+            _ => self
+                .custom_seq_binding(rdr, wrt, &mut evt, n, positive)?
+                .unwrap_or(Cmd::Unknown),
         })
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -519,6 +519,13 @@ fn readline_edit<H: Helper>(
             continue;
         }
 
+        #[cfg(windows)]
+        if cmd == Cmd::PasteFromClipboard {
+            use crate::tty::RawReader;
+            let clipboard = rdr.read_pasted_text()?;
+            s.edit_yank(&input_state, &clipboard[..], Anchor::Before, 1)?;
+        }
+
         // Tiny test quirk
         #[cfg(test)]
         if matches!(
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,8 @@ skim = { version = "0.9", optional = true }
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3", features = ["consoleapi", "handleapi", "minwindef", "processenv", "winbase", "wincon", "winuser"] }
 scopeguard = "1.1"
+clipboard-win = "4.1.0"
+error-code = "2.2.0"
 
 [dev-dependencies]
 doc-comment = "0.3"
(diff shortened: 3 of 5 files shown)


marcbowes:
> fd-lock rollback should be done in a dedicated PR after investigation

Sure thing!

> Maybe we should put clipboard-win behind a feature

What's the reasoning behind that? IMO, copy-paste should work out the box. Happy to make the change if you want, though. Let me know what the feature flag should be.

> For me, QuotedInsert should not be used but a new dedicated Cmd should be added.

Ok. What do you want it to be called?

marcbowes:
Pushed 2 changes: 

* revert the revert
* added `cmd::ClipboardPaste`. happy to accept new names

Note that C-q or C-v was used for QuotedInsert before. Now, I've bound C-v to ClipboardPaste by default on Windows. As far as I can tell, C-q doesn't really do anything my terminals on Windows anyways, but this is the smallest change I could make.

marcbowes:
I'm happy to squash commits once you're happy with the work.

gwenn:
"paste-from-clipboard" is the name used by [readline](https://git.savannah.gnu.org/cgit/readline.git/tree/funmap.c#n123).
So `PasteFromClipboard` would be perfect but `ClipboardPaste` is fine...

gwenn:
> Note that C-q or C-v was used for QuotedInsert before. Now, I've bound C-v to ClipboardPaste by default on Windows. As far as I can tell, C-q doesn't really do anything my terminals on Windows anyways, but this is the smallest change I could make.

Don't worry because most of the time `Ctrl-Q` does nothing on unix platform because by default a signal is sent instead.

src/keymap.rs
```diff
@@ -1046,9 +1049,13 @@ impl InputState {
                 } else {
                     Cmd::Kill(Movement::EndOfLine)
                 }
-            },
-            E(K::Char('Q'), M::CTRL) | // most terminals override Ctrl+Q to resume execution
+            }
+            E(K::Char('Q'), M::CTRL) => Cmd::QuotedInsert,
+            // most terminals override Ctrl+Q to resume execution
```
gwenn, line 1054: Please move this comment above `Ctrl-Q`...

src/lib.rs
```diff
@@ -519,6 +519,13 @@ fn readline_edit<H: Helper>(
             continue;
         }
 
+        if cmd == Cmd::PasteFromClipboard {
+            // Quoted insert
```
gwenn, line 524: Remove this comment.

src/keymap.rs
```diff
@@ -28,6 +28,9 @@ pub enum Cmd {
     CapitalizeWord,
     /// clear-screen
     ClearScreen,
+    /// Paste from the clipboard
+    #[cfg(windows)]
```
gwenn, line 32: Please, remove `#[cfg(windows]` and use `PasteFromClipboard` [here](https://github.com/kkawakam/rustyline/blob/master/src/keymap.rs#L1070-L1071)

src/lib.rs
```diff
@@ -519,6 +519,13 @@ fn readline_edit<H: Helper>(
             continue;
         }
 
+        #[cfg(windows)]
```
gwenn, line 522: Please, remove `#[cfg(windows]`

src/keymap.rs
marcbowes: > Please, remove `#[cfg(windows]`

You sure about that? I don't want to accidentally break other platforms. Further, the link you gave earlier (https://git.savannah.gnu.org/cgit/readline.git/tree/funmap.c#n123) also had a conditional on windows:

```
#if defined (_WIN32)
  { "paste-from-clipboard", rl_paste_from_clipboard },
#endif
```

> and use PasteFromClipboard here

ok. As an "or", right?

marcbowes: I pushed a commit that does this. I'd like to test first on Linux and macOS before merging. I can't do that today - it's already late in my tz :).

gwenn: You are right: if a unix user wants to bind a specific sequence to `PasteFromClipboard`, this would not work because `read_pasted_text` works only in bracketed paste mode...

marcbowes: Should I revert https://github.com/kkawakam/rustyline/pull/521/commits/7055c27ff2920f5a2e29cd902b791d2f2fdd3dcf on my branch then? I'm pretty comfortable with the level of testing and "no breakage of other platforms" as of that commit.

gwenn: Yes, please, revert 7055c27.

marcbowes:
After reverting https://github.com/kkawakam/rustyline/commit/7055c27ff2920f5a2e29cd902b791d2f2fdd3dcf again, I've tested that rustyline compiles and passes all tests on a Windows and macOS machine. I've tested that paste is fixed on my Windows machine and not affected on the macOS machine.

marcbowes:
I see a previous action run failed because of cargo-fmt failures (https://github.com/kkawakam/rustyline/runs/2533459886?check_suite_focus=true). Those are unrelated to my change. I can fix them if you'd like (I've actually been *undoing them* as I worked to avoid a noisy diff).

gwenn:
> I see a previous action run failed because of cargo-fmt failures (https://github.com/kkawakam/rustyline/runs/2533459886?check_suite_focus=true). Those are unrelated to my change. I can fix them if you'd like (I've actually been _undoing them_ as I worked to avoid a noisy diff).

It may be related to rust 1.52.0: `cargo fmt` may have changed.

marcbowes:
I think this is ready to go, but please approve the workflows to validate.

gwenn:
Thanks.

marcbowes:
Thanks for helping me with this. Do you have any idea when the next release (to crates.io) will be?

gwenn:
Nope.

link2xt:
@marcbowes This PR pulled in `str-buf` with this issue: https://github.com/DoumanAsh/str-buf/issues/1
It means `rustyline` now requires at least rust 1.51.0 to build. Maybe indeed put this behind the feature if you want to support older rust versions.

marcbowes:
Oh, that sucks. I'm not the maintainer of this library, so I'm tagging @gwenn for input.

I guess the options are:

1. Do nothing
2. Revert this commit and release 8.2.0 and/or:
3. Re-add this commit either as a new MV (9.0.0) or behind a feature flag

Again, I'm not the maintainer, but if it were up to me I'd revert this commit, release 8.2.0, then readd the commit and release as 9.0.0. IMO, this was an important bugfix and consumers should not have to turn on a feature flag to disable bugs.

link2xt:
Not sure it's important, but CI failed with rust 1.50.0 at https://github.com/deltachat/deltachat-core-rust/pull/2448
There was one request to keep supporting Debian version of rustc, so I will not upgrade for now.

Probably makes sense to pin rust version in GitHub Actions workflow of rustyline or test across several versions to avoid such surprises later.

DoumanAsh:
Sorry for this.
I lowered compiler requirement in `clipboard-win` and `error-code` and yanked versions where bump was introduced.
Do cargo update and see if it works

gwenn:
8.2.0 released.

PR #523, opened 2021-05-15
Fix line wrap predicate

flappyBug:
Raw command line ends with `\n`, but after highlighting, it appends things like `\x1B[0m` to the end. And this makes rustyline thinks it should add a extra `\n` to the end of buffer. See https://github.com/nushell/nushell/issues/3431. 

Since I'm not familiar with windows console, I didn't check if windows has the same problem.

### DIFF
--- a/src/tty/unix.rs
+++ b/src/tty/unix.rs
@@ -865,7 +865,12 @@ impl Renderer for PosixRenderer {
             }
         }
         // we have to generate our own newline on line wrap
-        if end_pos.col == 0 && end_pos.row > 0 && !ends_with_new_line(&self.buffer) {
+        if end_pos.col == 0
+            && end_pos.row > 0
+            && !hint
+                .map(|h| h.ends_with('\n'))
+                .unwrap_or_else(|| line.ends_with('\n'))
+        {
             self.buffer.push('\n');
         }
         // position the cursor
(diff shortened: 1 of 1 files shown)


src/tty/unix.rs
```diff
@@ -865,7 +865,7 @@ impl Renderer for PosixRenderer {
             }
         }
         // we have to generate our own newline on line wrap
-        if end_pos.col == 0 && end_pos.row > 0 && !self.buffer.ends_with('\n') {
+        if end_pos.col == 0 && end_pos.row > 0 && !line.ends_with('\n') {
```
gwenn, line 868: We need to handle the case where there is an hint.

flappyBug: Thank you for pointing out that. Do you have any suggestion? Another solution in my mind is to reverse search buffer, and skip escape sequence, take https://github.com/kkawakam/rustyline/blob/master/src/tty/mod.rs#L170 for reference.

src/tty/unix.rs
```diff
@@ -865,7 +865,7 @@ impl Renderer for PosixRenderer {
             }
         }
         // we have to generate our own newline on line wrap
-        if end_pos.col == 0 && end_pos.row > 0 && !self.buffer.ends_with('\n') {
+        if end_pos.col == 0 && end_pos.row > 0 && !ends_with_new_line(&self.buffer) {
```
gwenn, line 868: Why not simply do:
```rust
        if end_pos.col == 0
            && end_pos.row > 0
            && !hint
                .map(|h| h.ends_with('\n'))
                .unwrap_or_else(|| line.ends_with('\n'))
        {
```
?

flappyBug: Thank you. That's much simpler.

gwenn:
Version 8.1.0 released

PR #526, opened 2021-05-19
Remove need for error-code direct usage

DoumanAsh:
As follow up to https://github.com/kkawakam/rustyline/pull/521#issuecomment-844457084 I removed need for user to directly pull `error-code`


gwenn:
Thanks.

Issue #302, opened 2019-11-13
Feature Request: Use tabs for switching between auto completion values


oddcoder:
I would love to see tab  to switch between autocompleting options. something similar to what ZSH does like in this ascinema

[![asciicast](https://asciinema.org/a/00Efj9eOFHE2RThH7Um7RwkZQ.svg)](https://asciinema.org/a/00Efj9eOFHE2RThH7Um7RwkZQ)

I am not quite sure if this feature exists, but hidden somewhere, or it is not here yet.

gwenn:
[CompletionType::Circular](https://docs.rs/rustyline/5.0.4/rustyline/config/enum.CompletionType.html#variant.Circular) ?

oddcoder:
It is very similar it does the circulation but it doesn't print the elements that it circulates around.

jackgerrits:
This is similar to completion in zsh, or it might be a plugin brought in by Oh-My-Zsh? Would be an awesome feature

oddcoder:
I think it is plugin bought by ohmyzsh Still it is not possible to do that in rustyline is it?

gwenn:
Maybe we should enhance the `Completer` trait and let implementations do their own [complete_line](https://github.com/kkawakam/rustyline/blob/master/src/lib.rs#L67) logic.
But currently, the [Renderer](https://github.com/kkawakam/rustyline/blob/master/src/tty/mod.rs#L30) is very limited.

PR #529, opened 2021-05-29
feat: adding CircularList completion behavior

ahkrr:
This PR is adding CircularList behavior in the complete_line function.
The list will stay and update in place. 
Resizing the terminal can cause the displayed list to appear problematically, which can be resolved by another TAB, which will clear the terminal from the cursor-position until the end and redraw the list.
I wrote this because nushell is using rustyline and I wanted this feature for it.
I currently use oh-my-zsh.
This PR is also related to this #302
The gif below is nushell using the additions to rustyline.
t-rec

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -84,7 +84,7 @@ fn main() -> rustyline::Result<()> {
     env_logger::init();
     let config = Config::builder()
         .history_ignore_space(true)
-        .completion_type(CompletionType::List)
+        .completion_type(CompletionType::CircularList)
         .edit_mode(EditMode::Emacs)
         .output_stream(OutputStreamType::Stdout)
         .build();
--- a/src/completion_type.rs
+++ b/src/completion_type.rs
@@ -1,0 +1,426 @@
+pub use crate::binding::{ConditionalEventHandler, Event, EventContext, EventHandler};
+use crate::completion::{longest_common_prefix, Candidate, Completer};
+pub use crate::config::{
+    ColorMode, CompletionType, Config, EditMode, HistoryDuplicates, OutputStreamType,
+};
+use crate::edit::State;
+use crate::error;
+pub use crate::keymap::{Anchor, At, CharSearch, Cmd, InputMode, Movement, RepeatCount, Word};
+use crate::keymap::{InputState, Refresher};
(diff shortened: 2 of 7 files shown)


gwenn:
Thanks.
I would prefer that `Completer` trait is enhanced and let users implement their own `show_candidates` method.
Which means that we need an API to move cursor and write on screen.

ahkrr:
I only know as much about writing to a terminal as was needed for this PR.
In the rust ecosystem, is it really necessary to build another `API` to interact with terminals. 
Couldn't the `Completer` be extended without building the terminal `API` and having the users use things like [crossterm](https://lib.rs/crates/crossterm) or [console](https://lib.rs/crates/console).
Instead of building an API in rustyline, one could build examples, that work with [crossterm](https://lib.rs/crates/crossterm) or [console](https://lib.rs/crates/console) to show of some functionality.
Is this PR not of interest because of plans for rustyline or could it be integrated under certain circumstances?
Extensibility is a nice thing but having some choice in defaults is also nice.

gwenn:
Currently the [Renderer](https://github.com/kkawakam/rustyline/blob/master/src/tty/mod.rs#L30) trait is not public so only `rustyline` can write on terminal screen.
And I guess that `crossterm` and `console` expect exclusive access to terminal screen.
So using `crossterm` for completions means implementing [Term](https://github.com/kkawakam/rustyline/blob/master/src/tty/mod.rs#L201) with `crossterm`...

See how `linefeed` allows users to [write](https://docs.rs/linefeed/0.6.0/linefeed/prompter/struct.Prompter.html#method.writer_append) on screen (but not from [Completer](https://docs.rs/linefeed/0.6.0/linefeed/complete/trait.Completer.html) because of immutable ref).

ahkrr:
The `terminal` in `windows` or `unixes` isn't limited by ownership, is it? 
If someone were to use `crosserterm` etc. to manipulate the `terminal`, as long as they restore the position to what `rustyline` expects, `rustyline` should be able to draw itself without problems, should it not? If a terminal API would be included in `rustyline`, wouldn't it use the same ways to manipulate the terminal as `crossterm` etc. ?

gwenn:
Only if `crossterm` does not keep a [lock](https://doc.rust-lang.org/std/io/struct.StdoutLock.html) on stdout.

Issue #303, opened 2019-11-19
Add support for inputrc configuration


Timidger:
The readline library supports a [configuration file that initializes the library based on a user's defaults in their home director](https://www.gnu.org/software/bash/manual/html_node/Readline-Init-File.html). I'm assuming this feature is in scope for this project?

gwenn:
Yes, this should be possible.
* configuring settings should be easy,
* configuring key bindings should also be easy except that currently escape sequences (on unix platform) are hardcoded.
* supporting conditional init constructs, I am not sure.

[linefeed](https://github.com/murarth/linefeed/blob/master/src/inputrc.rs) already supports parsing `inputrc`

thomcc:
I wonder if something like `~/.config/rustyline/config.toml` (or whatever) wouldn't be better/simpler.

There's definitely value in being able to configure key bindings and other settings (there's a lot of heuristics that go on for guessing what a terminal actually supports vs what it reports), but... it seems like inputrc is very specific to readline and how it works.

We could probably hack custom settings into it (possibly abusing `$if rustyline` or something), and also ignore the parts we don't want to support... but it's not really ideal either way -- it's nice for someone who has their inputrc configured, but pretty awful if for someone coming in the other direction.

gwenn:
Just for reference:
https://docs.helix-editor.com/remapping.html

414owen:
@thomcc supporting `~/.config/rustyline/config.toml` (instead of `.inputrc`) means people would need to configure applications using rustyline differently from readline applications.

This was the approach taken by haskeline. I really dislike having to maintain different formats of the same config.

Having a rustyline-specific config for options specific to rustyline might not be a bad shout, but `.inputrc` should definitely be taken into account.

PR #533, opened 2021-06-07
Allow ctrl+{n,p} to change lines in emacs mode

### LINKED ISSUE #343: Ctrl-n and Ctrl-p should not be exclusively history navigation (in emacs mode)
Right now Ctrl-N and Ctrl-P are exclusively history operations:

```rust
KeyPress::Ctrl('N') => Cmd::NextHistory,
KeyPress::Ctrl('P') => Cmd::PreviousHistory,
```

They should probably just behave the same as down and up, which currently are:

```rust
KeyPress::Down => Cmd::LineDownOrNextHistory,
KeyPress::Up => Cmd::LineUpOrPreviousHistory,
```

As it is differs from the behavior in fish-shell, zsh, emacs, and probably others (as well as being kind of annoying to anybody who uses these heavily as I do).

(In `bash` the behavior of ctrl-n and ctrl-p is always history navigation, but the behavior of up and down is also always history navigation, e.g. it's another example where the behavior of arrow keys and ctrl-n/p match).

I can submit a PR if it would be accepted.

thomcc:
This switches ctrl+n and ctrl+p from exclusively history navigation to
history-or-line navigation.

This is consistent with the behavior of everything else I've tried
including emacs, zsh, fish, and probably more. And it makes these
shortcuts equivalent to up and down arrow, which is the behavior I've
found in absolutely everything that supports them (including
applications with no notion of histroy, like text editors and macOS
input text boxes).

At least IMO, these being equivalent to arrow keys is the "obvious"
behavior for these keybindings.

Additionally, it adds ctrl+shift+n and ctrl+shift+p supporting the old
behavior, as [this comment][issue342c2] implies that maybe it would be
useful to have "only history". I haven't seen this in anything else, but
it seems harmless.

Fixes #343

[issue342c2]: https://github.com/kkawakam/rustyline/issues/343#issuecomment-621153806

### DIFF
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -595,8 +595,10 @@ impl InputState {
                 }
             }
             E(K::Char('L'), M::CTRL) => Cmd::ClearScreen,
-            E(K::Char('N'), M::CTRL) => Cmd::NextHistory,
-            E(K::Char('P'), M::CTRL) => Cmd::PreviousHistory,
+            E(K::Char('N'), M::CTRL_SHIFT) => Cmd::NextHistory,
+            E(K::Char('P'), M::CTRL_SHIFT) => Cmd::PreviousHistory,
+            E(K::Char('N'), M::CTRL) => Cmd::LineDownOrNextHistory(1),
+            E(K::Char('P'), M::CTRL) => Cmd::LineUpOrPreviousHistory(1),
             E(K::Char('X'), M::CTRL) => {
                 if let Some(cmd) = self.custom_seq_binding(rdr, wrt, &mut evt, n, positive)? {
                     cmd


gwenn:
For `rustyline`, emacs mode means `readline` emacs mode which may not behave like `emacs`.
And I am almost sure that you can input multilines with `readline` by using Ctrl-V Ctrl-J.

thomcc:
Mmh, I mean, I disagree that as-is this is sensible for multiline use, or consistent with the other keybindings. I also don't really see C-v C-j as equivalent at all to real multiline support, since it doesn't really work with anything else. But I guess we just disagree here, and its you're project, so your call.

That said, as it is this is... quite annoying for me as a long-time emacs user, and I at least would like to solve it for myself.

Given that, would you accept patches implementing support for supporting some kind of configuration, probably inputrc (e.g. #303)?

gwenn:
A `nocompatible` feature flag may be the easiest way.
But a configuration file would be fine too.

gwenn:
- [ ] find differences between bash / readline and zsh / zle
- [ ] find differences between readline and python-prompt-toolkit
- [ ] find all commands impacted by multiline

PR #547, opened 2021-07-15
Keep the distinction between backspace and ctrl-h on windows

### LINKED ISSUE #543: how to apply key bindings
What's the secret sauce to get either of these to work in the `custom_key_bindings` example in Windows Terminal? Am I specifying the handler wrong?

```rust
    // rl.bind_sequence(
    //     KeyEvent {
    //         0: KeyCode::Char('H'),
    //         1: Modifiers::CTRL,
    //     },
    //     Cmd::Kill(Movement::BackwardChar(1)),
    // );

    rl.bind_sequence(KeyEvent::ctrl('h'), Cmd::Kill(Movement::BackwardChar(1)));
```

gwenn:
backspace vs ctrl-h
enter vs ctrl-m
tab vs ctrl-i

should fix #543

### DIFF
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -1060,9 +1060,11 @@ impl InputState {
             } else {
                 Movement::BackwardChar(n)
             }),
-            E(K::Char('J'), M::CTRL) | E::ENTER => Cmd::AcceptOrInsertLine {
-                accept_in_the_middle: true,
-            },
+            E(K::Char('J'), M::CTRL) | E(K::Char('M'), M::CTRL) | E::ENTER => {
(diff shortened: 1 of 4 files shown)


fdncred:
Thanks @gwenn. I tested this with slight changes to the custom_key_bindings example and Ctrl+H works great now!

gwenn:
> Thanks @gwenn. I tested this with slight changes to the custom_key_bindings example and Ctrl+H works great now!

You should be aware that this binding will only work on windows because on most unix Ctrl-H cannot be distinguish from backspace key...

PR #552, opened 2021-07-21
Allowing displaying a prompt during incomplete validation

ejose19:
This change will let consumers write a custom prompt on incomplete validation result, one of the use cases is allowing REPL to display an indicator that it's expecting more input before evaluating (like `node` multi line  displaying `... ` as pseudo-prompt).

Semver major due the breaking change.

### DIFF
--- a/examples/input_validation.rs
+++ b/examples/input_validation.rs
@@ -12,7 +12,7 @@ impl Validator for InputValidator {
         let result = if !input.starts_with("SELECT") {
             Invalid(Some(" --< Expect: SELECT stmt".to_owned()))
         } else if !input.ends_with(';') {
-            Incomplete
+            Incomplete(None)
         } else {
             Valid(None)
         };
--- a/src/edit.rs
+++ b/src/edit.rs
@@ -208,7 +208,11 @@ impl<'out, 'prompt, H: Helper> State<'out, 'prompt, H> {
             let result = validator.validate(&mut ValidationContext::new(self))?;
             let corrected = self.changes.borrow_mut().end();
             match result {
-                ValidationResult::Incomplete => {}
+                ValidationResult::Incomplete(ref msg) => {
+                    if msg.is_some() {
+                        self.refresh_line_with_msg(msg.as_deref())?;
+                    }
+                }
                 ValidationResult::Valid(ref msg) => {
                     // Accept the line regardless of where the cursor is.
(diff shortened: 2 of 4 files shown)


ejose19:
I see that:
https://github.com/kkawakam/rustyline/blob/250c9e6c6655de8f4b392e7196f618b5d8b06a95/src/command.rs#L145-L147

occurs after the layout change, so even if newline was inserted when there is a message, it won't be seen. Should this be refactored instead to be an option? (ie: `incompletePrompt`), as current implementation either inserts a new line or show the validation message, not both.

gwenn:
> This change will let consumers write a custom prompt on incomplete validation result, ...

You mean a custom hint ?

gwenn:
> as current implementation either inserts a new line or show the validation message, not both.

Yep currently you can't have both.

ejose19:
> > This change will let consumers write a custom prompt on incomplete validation result, ...
> 
> You mean a custom hint ?

I don't think this would clasify as a hint, since it will appear to the left of user input, and shouldn't disappear once user start typing nor when he enter more lines, ie (`\n` represents user triggered newlines, and `... ` would be the "pseudo-prompt" that appears on each incomplete line):

```
> {\n
... a:1\n
... }\n
{ a: 1 }
>
```

gwenn:
> I don't think this would clasify as a hint, since it will appear to the left of user input

`refresh_line_with_msg` is used to display a message / hint / suggestion...

ejose19:
> > I don't think this would clasify as a hint, since it will appear to the left of user input
> 
> `refresh_line_with_msg` is used to display a message / hint / suggestion...

Yes, I've tried with that, but there's 2 issues:

1. Input is placed in the same place as the message, given message being `... `:

```
> \n
...
^input is placed in this column (1)
    ^ but should be placed here (5)
```

2. As soon the user types something, the message disappear, but the desired behavior is for it to persist, even if new lines are added, the only time it should be removed if the user deletes that line.

gwenn:
What you want is a [continuation prompt](https://github.com/kkawakam/rustyline/pull/372).
https://python-prompt-toolkit.readthedocs.io/en/master/pages/asking_for_input.html?highlight=continuation#multiline-input

ejose19:
@gwenn The only limitation I see is that both prompts must have the same width, which wouldn't apply for `> ` & `... `, but it definitely works, thanks for the link.

PR #564, opened 2021-09-01
Fix crates.io badge

atouchet:


gwenn:
Thanks but [current](https://github.com/kkawakam/rustyline/blob/master/.github/workflows/rust.yml#L11-L14) rules prevent me from merging your PR (only the README file has been changed but Github Actions/builds are required).
But don't worry...

PR #574, opened 2021-11-12
Implement completion for executables in PATH

robiot:
This is a work in progress pull request. The completion works good, tested on windows and linux. Now behaves like kinda like bash.
Windows
Linux

Being able to configure if it should use PATH executable completion or not in the config builder would be nice to have, since I guess everyone doesn't want their readline to work like bash.

```rs
        .complete_env_executables(true) // Or false
        .build();
```
I have been trying to implement this, but with no success without making the code incredibly messy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ Cargo.lock
 
 # default history file
 history.txt
+
+testing/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,8 @@ cfg-if = "1.0"
 # For file completion
 # https://rustsec.org/advisories/RUSTSEC-2020-0053.html
 dirs-next = { version = "2.0", optional = true }
+# For PATH files completion
+searchpath = "0.1.3"
 # For History
 fd-lock = "3.0.0"
 libc = "0.2"
--- a/src/completion.rs
+++ b/src/completion.rs
@@ -1,7 +1,6 @@
 //! Completion API
 use searchpath::search_path;
 use std::borrow::Cow::{self, Borrowed, Owned};
-use std::ffi::OsString;
 use std::fs;
 use std::path::{self, Path};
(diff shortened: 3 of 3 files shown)


gwenn:
Command/path completion vs filename completion

Current home/`~` completion is put behind a cargo feature (`with-dirs`).
- [ ] command completion should be put behind a cargo feature (`searchpath` dependency should be optional)

- [ ] code related to command completion should be separated from code related to filename completion (`FilenameCompleter` should be kept untouched).

See how [bash_default_completion](https://github.com/bminor/bash/blob/master/bashline.c#L1791) and [command_word_completion_function](https://github.com/bminor/bash/blob/master/bashline.c#L2214) reuse `rl_filename_completion_function`.
Or see [CicadaCompleter](https://github.com/mitnk/cicada/blob/master/src/completers/mod.rs#L57) (`BinCompleter` vs `PathCompleter`).

Caller/user can then choose how to combine these two distinct completers.

robiot:
"Add completers" (ef9fe12)
Optional dependency: No
FilenameCompleter Untouched: Yes

This is not exactly what you asked for, and I am sorry.
Tho this gets the work done (in a kinda messy way, can be cleaned up).

You can now use BinCompleter, FilenameCompleter and ShellCompleter.
```rs
#[derive(Helper)]
struct MyHelper {
    completer: ShellCompleter,
    highlighter: MatchingBracketHighlighter,
    validator: MatchingBracketValidator,
    hinter: HistoryHinter,
    colored_prompt: String,
}
```
Where FilenameCompleter works the same as it did before (no changes), BinCompleter completes binaries in path, ShellCompleter does both in a shell way.

If you would want to have multiple "helpers" this is a plus.

robiot:
No updates in 7 months. Closing this for now.

PR #585, opened 2021-12-28
map C-f in Emacs mode to complete hints iff at end of line

joooooooooooooooooooooooooooooooooooosh:
This doesn't affect the behaviour of C-f if in the middle of a line, so I expect there'd be minimal unexpected impact.
This is just a QOL improvement that would be nice to have for C-f (even though the Right arrow key already does this), as popular programs like zsh already have this behaviour for C-f and there is very little current support for conditional/overloading mappings through the API for rustyline.

### DIFF
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -583,6 +583,8 @@ impl InputState {
                 Movement::ForwardChar(n)
             }),
             E(K::Char('E'), M::CTRL) => Cmd::Move(Movement::EndOfLine),
+            // Don't complete hints when the cursor is not at the end of a line
+            E(K::Char('F'), M::CTRL) if wrt.has_hint() && wrt.is_cursor_at_end() => Cmd::CompleteHint,
             E(K::Char('F'), M::CTRL) => Cmd::Move(if positive {
                 Movement::ForwardChar(n)
             } else {


gwenn:
https://github.com/kkawakam/rustyline/blob/master/examples/custom_key_bindings.rs#L52

joooooooooooooooooooooooooooooooooooosh:
I'm aware that this is doable (if a bit messy) using ConditionalEventHandler, but in this case since C-f wouldn't do anything normally when at the end of a line, is there a particular reason not to add extra behavior here? Given the Right arrow key already does this it feels natural to have C-f also be able to expand hints at the end of a line.

gwenn:
> I'm aware that this is doable (if a bit messy) using ConditionalEventHandler, but in this case since C-f wouldn't do anything normally when at the end of a line, is there a particular reason not to add extra behavior here? Given the Right arrow key already does this it feels natural to have C-f also be able to expand hints at the end of a line.

The Right arrow key behaviour was changed before `ConditionalEventHandler`.
See #533 for a similar discussion.
Should we
* introduce a zsh / zle mode,
* diverge further from emacs mode (as implicitly defined by `readline`) when principle of least surprise applies,
* ...
?

joooooooooooooooooooooooooooooooooooosh:
For both this PR and #533, I would still argue that the principle of least surprise applies more to having C-[npbf] match the behaviour of the relevant arrow keys *within the same application*.
I don't think making the behaviour of these keys diverge from one another simply because the behaviour of the arrow keys were changed before `ConditionalEventHandler` was added is a good idea.

If there are more differences between common zsh bindings and readline bindings, it would probably be worth creating a seperate zsh / zle mode, but either way I think it's worth trying to make arrow keys consistent with their C-[npfb] counterparts in both modes.

PR #587, opened 2021-12-29
Bind CTRL-D to Cmd::EndOfFile on Windows

### LINKED ISSUE #586: CTRL-D does not send EoF in Windows Terminal or Console Host
The following code from the _README_ does not trigger when running the example in _Windows Terminal_ or _Console Host_:

```rust
Err(ReadlineError::Eof) => {
    println!("CTRL-D");
    break
},
```

Pressing <kbd>CTRL</kbd>+<kbd>D</kbd> does not seem to result in any result at all (neither an empty line nor an error).

The <kbd>CTRL</kbd>+<kbd>C</kbd> combination seems to work as intended (`ReadlineError::Interrupted` is returned).

95jonpet:
- Return Cmd::EndOfFile when there is no input (nothing to kill in emacs mode).
- Fix invalid cfg! macro argument.

Fixes #586.

### DIFF
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -1068,7 +1068,7 @@ impl InputState {
                     } else {
                         Movement::BackwardChar(n)
                     })
-                } else if cfg!(window) || cfg!(test) || !wrt.line().is_empty() {
+                } else if cfg!(windows) || cfg!(test) || !wrt.line().is_empty() {
                     Cmd::EndOfFile
                 } else {
                     Cmd::Unknown


gwenn:
Thanks for the PR.
But I was wrong. Only @thomcc is right.
Could you please fix only `cfg!(windows)` and keep `!wrt.line().is_empty()` untouched.

src/keymap.rs
```diff
@@ -1068,7 +1068,7 @@ impl InputState {
                     } else {
                         Movement::BackwardChar(n)
                     })
-                } else if cfg!(window) || cfg!(test) || !wrt.line().is_empty() {
+                } else if cfg!(windows) || cfg!(test) || wrt.line().is_empty() {
```
gwenn, line 1071:
```rust
                } else if cfg!(windows) || cfg!(test) || !wrt.line().is_empty() {
```
should be good.

gwenn (changes requested):

gwenn (approved):

PR #604, opened 2022-02-23
Introduce History trait

### LINKED ISSUE #127: More flexible history save/load
I would like to be able to save user input history to and load it from a database. However, the current load/save API requires using a file. It *is* kind of possible to load/save history in line-by-line fashion, but this is suboptimal - there is no method on `History` to access the underlying `VecDeque`, and there is also no way to fill it with lines without using `add()` which performs the history addition logic.

One possible solution I see would be to implement the following pair of methods:
```rust
pub fn replace_entries<I, S>(&mut self, lines: I) where I: IntoIter<Item=S>, S: Into<String> {}
pub fn entries(&self) -> impl Iterator<Item=&str> {}
```
I can do this, but will the PR be merged? I see that the repo is not very active.

### LINKED ISSUE #442: Save lines in history file with a Unix timestamp
Just What I said above...

gwenn:
Fix #127
Fix #442

TODO
- [x] check if `History::add` and `History::add_owned` can be fused (https://users.rust-lang.org/t/trait-method-that-supports-both-str-and-string-as-parameter/72498 suggests that this is not possible)
- [x] put old / default `FileHistory` behind a feature
- [X] SQLite based implementation behind a feature
- [ ] timestamped entry ? Implementation detail ?
- [ ] editable entry ? later ?

Breaking changes:
* `Editor` impacted (new generic type parameter `I`)
* `History` renamed to `FileHistory`
* History API impacted (to make trait object safe)


gwenn:
If anyone is interested, I would like a code review.
Mainly on the [History](https://github.com/gwenn/rustyline/blob/history_trait/src/history.rs#L46-L175) trait.
And on the [SQLiteHistory](https://github.com/gwenn/rustyline/blob/history_trait/src/sqlite_history.rs) implementation.
Which differs from [reedline](https://github.com/nushell/reedline/blob/main/src/history/sqlite_backed.rs):
* all entries are kept in the SQLite database.
* and so entries may be added concurrently by different sessions.

Issue #542, opened 2021-07-13
Support for arbitrary streams for input/output


slashrsm:
It seems that `rustyline` currently only supports standard input/output. [Goose](https://github.com/tag1consulting/goose) will accept connections via telnet and expose a command line interface, which allows users to control their load tests. We would like to use `rustyline` to improve the UX of the said command line, but we would need to be able to wire it into a TCP stream instead of stdin/stdout. 

The idea of supporting arbitrary streams was already floated in #377. This approach would allow us to achieve what we want while also opening a lot of other possibilities for the library. 

Is this something that would be considered? Are there any plans to do that already?

gwenn:
Like https://github.com/jline/jline3/tree/master/remote-ssh/src/main/java/org/jline/builtins/ssh or https://github.com/prompt-toolkit/python-prompt-toolkit/tree/master/examples/ssh ?

slashrsm:
Yes, something like this.

gwenn:
I have no idea how jline3 / python-prompt-toolkit support ssh / telnet session.
Maybe only the [Terminal](https://github.com/kkawakam/rustyline/blob/master/src/tty/mod.rs#L201) implementation differs ?

slashrsm:
Possibly another implementation of the `Term` trait? Additionally to the Windows and Unix ones that we have already?

gwenn:
In this case, we need a cargo feature to activate the ssh / telnet implementation.

slashrsm:
Agreed. Are you generally prepared to accept this feature if I give it a try?

gwenn:
If only `tty` module is impacted and all new dependencies are optional except when this feature is activated, I am ok.

slashrsm:
OK, I will give it a try and see where it takes me. Thank you for the fast response!

m-mueller678:
I implemented this [here](https://github.com/m-mueller678/rustyline/) by adding a new variant to `config::Behaviour` that allows the user to specify file descriptors to use. Would you like me to file a PR?

PR #612, opened 2022-04-06
add support for arbitrary file descriptors on unix

m-mueller678:
This allows the use of arbitrary file descriptors on unix in response to issue #542. File Descriptors are supplied via a new variant of `config::Bhaviour`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,9 +65,14 @@ default = ["with-dirs"]
 with-dirs = ["dirs-next"]
 with-fuzzy = ["skim"]
 case_insensitive_history_search = ["regex"]
+arbitrary-file-descriptors=[]
 
 [package.metadata.docs.rs]
 features = ["with-dirs", "with-fuzzy"]
 all-features = false
 no-default-features = true
 default-target = "x86_64-unknown-linux-gnu"
+
+[[example]]
+name = "arbitrary_file_descriptors"
+required-features = ["arbitrary-file-descriptors"]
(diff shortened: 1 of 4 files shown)


gwenn:
Could you please explain why only unix is supported ?
And I am afraid that this PR doesn't work for telnet / ssh connection because `isatty` will return false ([jline3](https://github.com/jline/jline3/blob/master/remote-ssh/src/main/java/org/jline/builtins/ssh/ShellFactoryImpl.java#L115) seems to use the `system` flag to skip `isatty` check). So I guess we need something more complicated than a file descriptor...

m-mueller678:
My main use case for this feature is working with pseudoterminals other than the one controlling the process.

Only Unix is supported because I don't have a Windows machine to work on and don't have any experience with Windows terminal handling. Not for any technical reason.

This does indeed not work well with things other than TTYs. I don't think you can reliably support anything that is not a TTY because you cannot change to raw mode without termios as far as I know. Do you think a flag to override `isatty` and just assume the user set everything up as needed would be a good idea? If so, `Behavior::ArbitraryFileDescriptors` seems like a good place to put it. Anyone using this should have to consciously make this choice, and the other behaviours likely don't need it.

As a side note: My current workaround for supporting ssh is to just hijack the pseudo terminal slave. If necessary, I send SIGSTOP to its controlling process to ensure it doesn't steal any of my input.

eldoccc:
I assume this feature isn't in discussion anymore ? My use case is for telnet and while trying to play with telnet settings it became obvious that character mode(or linemode in EDIT) with overriden IOs wasn't enough to get stuff like tab or arrow keys working.

PR #613, opened 2022-04-10
Replace `buf_redux` with `BufReader` from `std`

lopopolo:
`rustyline` depends on `buf_redux` for two calls to `get_len`, but these
calls can be replaced with calls to `reader.buffer().len()` using APIs
in `std` that have been available since Rust 1.37.0:

https://doc.rust-lang.org/std/io/struct.BufReader.html#method.buffer

`std::io::BufReader::buffer` does not cause any additional reads to
occur and exposes the internal buffered and unread contents of the
`BufReader`.

Motivation for this commit is reducing dependencies.


gwenn:
Indeed, thanks.

PR #614, opened 2022-04-10
Explore removing `SmallVec` from `Event` enum

lopopolo:
This PR explores removing the dependency on `smallvec` and replacing it with `Vec` from `std`.

`smallvec` is a third party dependency that leaks into the public API. If applications that depend on `rustyline` wish to create custom key bindings, they must also depend on `smallvec` to construct `Event::KeySeq`.

This PR replaces `smallvec`, which is a complicated data structure with lots of unsafe code and several known security advisories, with a vocabulary type from `std`.

This has the effect of reducing the size of the `Event` enum.

Because `Event` is a public enum, changing the type of the field in `Event::KeySeq` is a breaking change. I've shown this change in the sequence of commits in this branch and the addition of a test.

Motivation of this PR is reducing dependencies.

@gwenn if you'd prefer, I'd be willing to try an alternate PR that conditionally enables `smallvec` with a Cargo feature. I think it looks easy enough to support.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,8 +36,6 @@ unicode-width = "0.1"
 unicode-segmentation = "1.0"
 memchr = "2.0"
 # For custom bindings
-# https://rustsec.org/advisories/RUSTSEC-2021-0003.html
-smallvec = "1.6.1"
 radix_trie = "0.2"
 regex = { version = "1.5.4", optional = true }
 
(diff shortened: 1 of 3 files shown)


gwenn:
```sh
$ cargo tree
...
├── radix_trie v0.2.1
│   ├── endian-type v0.1.2
│   └── nibble_vec v0.1.0
│       └── smallvec v1.8.0
├── smallvec v1.8.0
```

lopopolo:
😅 I am hoping to follow up with additional PRs that make the `radix_trie` dependency optional as well.

lopopolo:
> 😅 I am hoping to follow up with additional PRs that make the `radix_trie` dependency optional as well.

I've followed up on this here: https://github.com/kkawakam/rustyline/pull/615

PR #615, opened 2022-04-10
Make custom key bindings an optional Cargo feature

lopopolo:
Add a cargo feature called `custom-bindings` which is enabled by
default. The `custom-bindings` feature enables APIs which allow users of
`rustyline` to define custom keybindings.

Custom keybindings are the only use of `radix_trie` in this crate. If
the `custom-bindings` feature is disabled, `radix_tree` is not included
in `rustyline`.

Motivation of this commit is to reduce dependencies.

The approach I've taken in this commit is to stub out internal APIs when
`custom-bindings` feature is not active to minimize the number of code
changes that need to occur. Similarly, when the `custom-bindings`
feature is not enabled, I've added `allow(dead_code)` to `mod bindings`.

The splitting out of default functionality into a new feature that is
enabled by default I believe is a breaking change, because users may
have previously been using custom keybindings with default-features =
false.

This commit adds a step to the GitHub Actions CI to run `cargo check`
with no default features.


lopopolo:
Cargo tree output now looks like this:

```console
$ cargo tree -e no-dev --no-default-features
rustyline v9.1.2 (/Users/lopopolo/dev/repos/rustyline)
├── bitflags v1.3.2
├── cfg-if v1.0.0
├── fd-lock v3.0.5
│   ├── cfg-if v1.0.0
│   └── rustix v0.34.2
│       ├── bitflags v1.3.2
│       ├── errno v0.2.8
│       │   └── libc v0.2.122
│       ├── io-lifetimes v0.6.1
│       └── libc v0.2.122
├── libc v0.2.122
├── log v0.4.16
│   └── cfg-if v1.0.0
├── memchr v2.4.1
├── nix v0.23.1
│   ├── bitflags v1.3.2
│   ├── cfg-if v1.0.0
│   ├── libc v0.2.122
│   └── memoffset v0.6.5
│       [build-dependencies]
│       └── autocfg v1.1.0
├── unicode-segmentation v1.9.0
├── unicode-width v0.1.9
└── utf8parse v0.2.0
```

gwenn:
- [ ] https://github.com/kkawakam/rustyline/blob/master/Cargo.toml#L67
- [ ] `#![cfg_attr(docsrs, feature(doc_cfg))]`

lopopolo:
thanks for the review @gwenn. I took a stab at your feedback using a similar approach I take in some other crates I maintain.

When building the docs like this:

```sh
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --open
```

These methods on `Editor` render like this:

Screen Shot 2022-04-10 at 12 04 46 PM

src/keymap.rs
```diff
@@ -1,11 +1,135 @@
...
 use crate::keys::{KeyCode as K, KeyEvent, KeyEvent as E, Modifiers as M};
 use crate::tty::{self, RawReader, Term, Terminal};
-use crate::{Config, EditMode, Event, EventContext, EventHandler};
+use crate::{Config, EditMode, Event};
+#[cfg(feature = "custom-bindings")]
+use crate::{EventContext, EventHandler};
+use log::debug;
+
+#[cfg(feature = "custom-bindings")]
+mod custom_bindings {
```
lopopolo, line 12: I'm not sure if you'd prefer for me to use `cfg_if::cfg_if!` here

gwenn:
https://github.com/lopopolo/rustyline/pull/2

lopopolo:
Oh sorry @gwenn I missed the notification on my fork

PR #623, opened 2022-04-21
Upgrade nix to 0.24, limit features

rtzoeller:
This reduces compilation time slightly, and removes the memoffset crate as an indirect dependency.


gwenn:
Thanks

PR #625, opened 2022-04-22
Use lookup tables instead of linear search with `memchr`

lopopolo:
Several places use `memchr` with a byteset. This commit refactors these
code paths to construct a lookup table from `u8` -> `bool`, where an
index is set to `true` if the byte is present in the given slice.

This change removes linear scans that occur in loops, which changes
these functions runtime complexity from `O(m * n)` to `O(m + n)`.

### DIFF
--- a/src/completion.rs
+++ b/src/completion.rs
@@ -5,7 +5,6 @@ use std::path::{self, Path};
 
 use crate::line_buffer::LineBuffer;
 use crate::{Context, Result};
-use memchr::memchr;
 
 /// A completion candidate.
 pub trait Candidate {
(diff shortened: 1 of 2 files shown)


gwenn:
I guess `break_chars_byteset` can be evaluated once / statically.
Something like https://github.com/sqlite/sqlite/blob/master/src/tokenize.c#L61-L80 but for `DEFAULT_BREAK_CHARS` and `DOUBLE_QUOTES_SPECIAL_CHARS`.

lopopolo:
@gwenn I think so too but I wasn't sure if that change would be accepted since these functions are public APIs that take the byeset as a slice.

I'm not sure how you'd like to make the API breaks.

Do you want to merge this as is or maybe push a PR to my fork?

gwenn:
See #676

lopopolo:
Thanks @gwenn!

PR #638, opened 2022-06-17
Update custom key bindings example

sigmaSd:
In my testing selfinsert doesn't seem to work, it inserts one white-space character instead of an actual tab, changing it to insert makes it work perfectly.


sigmaSd:
I found that this example didn't work on windows, I added a fix but there is still a bug in rustyline where inserting a tab in windows inserts a tab but the cursor doesn't move, I added a workaround for it (inserting 4 spaces).

PR #639, opened 2022-07-08
Add diy command completor example

0xpause:
Add an example for diy command completor.

### DIFF
--- a/examples/diy_completor.rs
+++ b/examples/diy_completor.rs
@@ -1,0 +1,144 @@
+use rustyline::completion::{extract_word, Completer, FilenameCompleter, Pair};
+use rustyline::error::ReadlineError;
+use rustyline::hint::Hinter;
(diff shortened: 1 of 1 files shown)


examples/diy_completor.rs
```diff
@@ -0,0 +1,144 @@
...
+                } else {
+                    Ok((start, matches))
+                }
+            },
+            Err(e) => Err(e),
+        }
+    }
+}
+
+impl Hinter for MyHelper {
```
gwenn, line 92: you should derive `Hinter` above and remove this block.

examples/diy_completor.rs
```diff
@@ -0,0 +1,144 @@
...
+                } else {
+                    None
+                }
+            })
+            .collect();
+        Ok((start, matches))
+    }
+}
+
+impl Completer for CommandCompleter {
```
gwenn, line 50: Not strictly necessary.
You should call `find_matches` directly.

PR #670, opened 2023-01-16
Add derive feature that reexports rustyline-derive's API

figsoda:


gwenn:
Check:
- [ ] (cargo doc)[https://github.com/kkawakam/rustyline/blob/master/Cargo.toml#L88]
- [ ] (doc.rs)[https://github.com/kkawakam/rustyline/blob/master/Cargo.toml#L88] features
- [ ] (test/example)[https://github.com/kkawakam/rustyline/blob/master/Cargo.toml#L60]

figsoda:
> cargo doc

not sure what documentation to add since there wasn't any for rustyline-derive, but it does show up in the `Derive macros` section on docs.rs

image

gwenn:
> > cargo doc
> 
> not sure what documentation to add since there wasn't any for rustyline-derive, but it does show up in the `Derive macros` section on docs.rs 

I am not sure if it is possible but ideally they should be marked as available only with `derive` feature like [here](https://docs.rs/rustyline/latest/rustyline/struct.EventContext.html).

figsoda:
that seems to be possible, I have added that

gwenn:
Thanks !

PR #686, opened 2023-03-05
Make MemHistory::default() be a wrapper around MemHistory::new() to avoid that max_len is defaulted to 0.

tage64:
I found that when creating `MemHistory` from `Default::default()`, the `max_len` (that is the maximum size of the history) is initialized to `0` due to the `#[derive(Default)]` on the `MemHistory` struct.
This is counter intuative since the history doesn't work as a history at all in that case.

I changed the `Default` implementation to call `MemHistory::new()` instead. Then the history max length will be initialized to `100` instead.

### DIFF
--- a/src/history.rs
+++ b/src/history.rs
@@ -175,7 +175,6 @@ pub trait History {
 }
 
 /// Transient in-memory history implementation.
-#[derive(Default)]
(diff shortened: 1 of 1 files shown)


gwenn:
Thanks.

Issue #598, opened 2022-02-08
Indentation prompt for multiline edits


RubixDev:
I am currently working on creating a REPL for my own language. For multiline editing I just use the matching bracket validator for now, but that puts the cursor to the very front of the next line:
```
>> fun greet(name) {
    printl('Hello, ' + name + '!')
}
>> 
```
Comparing this to other REPLs like node or bpython, I would want it to look like this:
```
>> fun greet(name) {
..     printl('Hello, ' + name + '!')
.. }
>> 
```
Auto-indentation is not needed, just the placeholder dots in front to align the next lines with the prompt.

I think this would fit best as a config setting like `indent_text` which is `""` by default.

If there already is a way to achieve this, please let me know

RubixDev:
Are there any plans on merging that PR? reedline seems a bit unfinished and less mature to me and I dont't really want to switch the whole library for that one feature

gwenn:
Not in its current state.

Issue #417, opened 2020-07-18
Changing the prompt string dynamically


SirWindfield:
Hey!

I have a rather "special" use case. I want the user to enter a string of a specific length `n`. For convenience I wanted to show the remaining characters that they have while typing. Something like this, with `>>` being the prompt:
```
(56) >> User already typed stuff
```

Is this somehow possible?

gwenn:
No, you cannot but you may use `highlight_pompt` as a workaround with some limitations:
https://github.com/kkawakam/rustyline/pull/369#issuecomment-616720954

gwenn:
For reference, `replxx` uses a callback:
https://github.com/AmokHuginnsson/replxx/blob/master/examples/cxx-api.cxx#L228-L232

PR #696, opened 2023-04-02
Generic Prompt Type

hoseinzadeh:
Sometimes we want to asynchronously update the prompt. For example, if we are using rustyline as a remote terminal, we want to display the connection info in the prompt. When disconnected, it is nicer to show a message and change the prompt to a ready state. With a Generic prompt type, it is possible to do that.

Performance overhead: in order to display the prompt, it should implement ToString trait. Therefore, it has a little performance overhead because it calls to_string every time it needs the prompt as a string (printing the prompt or calculating the cursor position). However, it is negligible as the whole system is a human interface that does not affect functionality.

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -14,9 +14,9 @@ pub enum Status {
     Submit,
 }
 
-pub fn execute<H: Helper>(
+pub fn execute<H: Helper, P: ToString>(
     cmd: Cmd,
-    s: &mut State<'_, '_, H>,
+    s: &mut State<'_, H, P>,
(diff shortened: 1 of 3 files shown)


gwenn:
Prompt related stuff:
#208
https://github.com/kkawakam/rustyline/pull/372
https://github.com/kkawakam/rustyline/pull/369
#417
#598
https://github.com/kkawakam/rustyline/commit/4ec26e824e0f94b7f8f763d201093052805770c6

PR #215, opened 2019-05-10
Fix Windows width calculation

jonathandturner:
This fixes the Windows prompt width calculation by using the same width calculation Unix uses. This code might have originally been in place to work around a Windows shortcoming, but it seems that current versions of Windows calculate width the same way Unix does

Before:

beforefix

After:

afterfix

PR #702, opened 2023-05-31
Fix position calculation on Windows

ShaneEverittM:
The implementations for Renderer::calculate_position differed slightly between unix and Windows. Notably, the Windows implementation did not use tty::width, but instead used UnicodeWidthStr::width. It seems that the latter did not correctly compute the length.

For context, this bug was discovered in the boa JS interpreter, where on Windows the prompt being bold and blue caused the cursor to be offset by 9 characters.

### DIFF
--- a/src/tty/windows.rs
+++ b/src/tty/windows.rs
@@ -12,7 +12,6 @@ use std::sync::Arc;
 
 use log::{debug, warn};
 use unicode_segmentation::UnicodeSegmentation;
-use unicode_width::UnicodeWidthStr;
 use winapi::shared::minwindef::{BOOL, DWORD, FALSE, TRUE, WORD};
 use winapi::shared::winerror;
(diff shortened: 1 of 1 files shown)


gwenn:
See https://github.com/kkawakam/rustyline/pull/215#issuecomment-491945339

ShaneEverittM:
Dang, not sure how I missed that when looking for duplicates! Sorry!

gwenn:
1. So either we introduce some code / dependencies to remove ascii escape sequences when colours are not supported or wanted (I would prefer not to).
2. Or you make sure that raw prompt / input contains no ascii escape sequences (and only your `Highlighter` adds some style).

But neither 1 or 2 are compatible with continuation prompt (see https://github.com/kkawakam/rustyline/pull/372#issuecomment-626295074).

PR #710, opened 2023-06-24
Add `HistoryHinter::default()` and `HistoryHinter::new()`

segeljakt:
This allows me to `#[derive(Default)]` when using a `HistoryHinter` (like the other provided rustyline utilities).


gwenn:
Thanks.

PR #722, opened 2023-08-14
Fixing hints when not editing at end of line

jerbs:
Currently, hints are only working correctly when editing at the end of the line.
Let me describe the current behaviour using a simple example:
The line contains "abcd" and the cursor is at position zero when typing "123". That text has the hint "1234567890_".

1) rustyline displays: 123|abcd{456789_}
2) After typing a 4: 1234|abcd{56789_}
3) After cursor right: 1234a|bcd{56789_}
4) After cursor right: 1234ab|cd{56789_}
5) After cursor right: 1234abc|d{56789_}
6) After cursor right: 1234abcd|{56789_}
7) After cursor right: 1234abcd56789_|

The pipe character indicates the cursor position. The curly braces do contain the hint.

Multiple issues:
- Hint is displayed at end of line, not directly right of cursor.
- Hint stays active when moving the cursor.
- Remaining part of hint is finally inserted at the end of the line.
- No possibility to complete a hint when not at end of line.

This pull request fixes all four issues.

### DIFF
--- a/src/edit.rs
+++ b/src/edit.rs
@@ -128,7 +128,8 @@ impl<'out, 'prompt, H: Helper> State<'out, 'prompt, H> {
         if self.layout.cursor == cursor {
             return Ok(());
         }
-        if self.highlight_char() {
+        if self.highlight_char() || self.has_hint() {
+            self.hint = None;
             let prompt_size = self.prompt_size;
             self.refresh(self.prompt, prompt_size, true, Info::NoHint)?;
         } else {
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -562,8 +562,8 @@ impl<'b> InputState<'b> {
                     Cmd::CompleteBackward
                 }
             }
-            // Don't complete hints when the cursor is not at the end of a line
             E(K::Right, M::NONE) if wrt.has_hint() && wrt.is_cursor_at_end() => Cmd::CompleteHint,
+            E(K::Right, M::ALT) if wrt.has_hint() => Cmd::CompleteHint,
             E(K::Char('K'), M::CTRL) => Cmd::Kill(if positive {
                 Movement::EndOfLine
             } else {
--- a/src/tty/unix.rs
+++ b/src/tty/unix.rs
@@ -979,11 +979,21 @@ impl Renderer for PosixRenderer {
         }
         // display hint
         if let Some(hint) = hint {
+            // position the cursor within the line
+            if cursor.col > 0 {
+                write!(self.buffer, "\r\x1b[{}C", cursor.col).unwrap();
+            } else {
+                self.buffer.push('\r');
+            }
+            // enter insert mode:
+            write!(self.buffer, "\x1b[4h").unwrap();
             if let Some(highlighter) = highlighter {
                 self.buffer.push_str(&highlighter.highlight_hint(hint));
             } else {
                 self.buffer.push_str(hint);
             }
+            // leave insert mode:
+            write!(self.buffer, "\x1b[4l").unwrap();
         }
         // we have to generate our own newline on line wrap
         if end_pos.col == 0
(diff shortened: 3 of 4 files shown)


gwenn:
Could you please provide your `Hinter` implementation ?
Or to say otherwise, are you sure that the issue is in rustyline and not in your hinter ?

src/edit.rs
```diff
@@ -128,7 +128,8 @@ impl<'out, 'prompt, H: Helper> State<'out, 'prompt, H> {
         if self.layout.cursor == cursor {
             return Ok(());
         }
-        if self.highlight_char() {
+        if self.highlight_char() || self.has_hint() {
```
gwenn, line 131: Why moving the cursor should invalidate the hint ?
At least, this should be configurable, not hardcoded.

src/keymap.rs
```diff
@@ -562,8 +562,8 @@ impl<'b> InputState<'b> {
                     Cmd::CompleteBackward
                 }
             }
-            // Don't complete hints when the cursor is not at the end of a line
             E(K::Right, M::NONE) if wrt.has_hint() && wrt.is_cursor_at_end() => Cmd::CompleteHint,
+            E(K::Right, M::ALT) if wrt.has_hint() => Cmd::CompleteHint,
```
gwenn, line 566: I would prefer not to have this default binding in `rustyline`.
See `ConditionalEventHandler` to add your own custom binding.

src/tty/unix.rs
```diff
@@ -979,11 +979,21 @@ impl Renderer for PosixRenderer {
         }
         // display hint
         if let Some(hint) = hint {
+            // position the cursor within the line
```
gwenn, line 982: So your PR works only on unix, right ?

jerbs:
> Could you please provide your `Hinter` implementation ?
> Or to say otherwise, are you sure that the issue is in rustyline and not in your hinter ?

See https://github.com/jerbs/fsidx for my hinter implementation.
The main branch uses to official rustyline. The hint branch uses the forked rustyline with my proposed fix.

src/edit.rs
jerbs: In my example in the initial description go the pull request the cursor is moved in steps (3) to (7). In all these positions it does not make sense to insert the remaining partial part of the hint. Try to use it with the app I uploaded https://github.com/jerbs/fsidx. It has a shell mode with hints and completions implemented for long options starting with "--". I don't see in which scenario the currently implemented behaviour is useful.

src/keymap.rs
jerbs: Yes, I have seen that API. I can remove that commit from the pull request, if that's your preference. Just thought that it gives a better out of the box experience for everybody.

src/tty/unix.rs
jerbs: I do not have access to any Windows. I also don't know if Windows terminals have an insert mode? The resulting difference between Unix and Windows is that for Windows the hint is still displayed at the end of the line while it is displayed right next to the cursor on Unix. So, the difference is on presentation level only.

src/keymap.rs
jerbs: Using Alt+Right to get the hint completed also feels quite natural to me. With the rustyline default Alt+Right jumps to the end of the word. And that's exactly what happens when the hint is displayed right next to the cursor and Alt+Right is used.

src/tty/unix.rs
gwenn: See https://github.com/kkawakam/rustyline#wine

src/keymap.rs
gwenn: See https://github.com/kkawakam/rustyline/issues/406 where Ctrl-E feels quite natural to someone else...

Issue #726, opened 2023-08-29
Is it possible for a Highlighter to re-render a line to undo cursor-based highlights?


luketpeterson:
Consider the HistoryHinter example here: https://github.com/kkawakam/rustyline/blob/master/examples/example.rs which uses the `MatchingBracketHighlighter`.

If you type `( x )`, it will bold-blue the first '(', as it should, when the closing ')' is typed.  But if you then hit `Enter`, the '(' remains bold-blued on the now-complete line.

I have tried to fix this with my own implementation of a Highligher object, but neither the `highlight_char` nor the `highlight` Highlighter methods are called again on that original line.  (They are called on a new line with a len=0, pos=0)

Is there a way (perhaps with the `Validator`?) to get the editor to re-render the line one last time?  I see people may not want this as the default behavior if their tty speed is limited, but is there a way to make it happen?

Thank you.

luketpeterson:
I was able to get the behavior I want by adding an additional "needs_final_refresh" method into `Highlighter`, that could cause this code path to be taken:

https://github.com/kkawakam/rustyline/blob/1c4091a0b20d68dda50c902abb723a6a74739ab9/src/edit.rs#L237C1-L238C1

Is there a way to do this in a more appropriate way?  Or what design would you like to see in a PR?

Thanks again.

gwenn:
We already refresh the line here:
https://github.com/kkawakam/rustyline/blob/1c4091a0b20d68dda50c902abb723a6a74739ab9/src/command.rs#L28-L32
But indeed currently, we cannot check / know that the `Highlighter` needs a final refresh.

luketpeterson:
If adding a method to the `Highlighter` trait is distasteful, perhaps `highlight_char` could be called on a line that has the final EOL character at the end?

gwenn:
No worry, we can add a new method to `Highlighter` trait with a default implementation.

luketpeterson:
I tried to tie up this logic nicely in a PR, and I added the below method to the `Highlighter` trait:

```Rust
fn final_highlight(&self, line: &str) -> bool
```

But I realized there is a complication.

The internal-mutability pattern (ie `Cell`) used in the `MatchingBracketHighlighter` means that the `highlight_char()` method actually has the side-effect of updating the highlighter state.

I followed this pattern and used the `final_highlight()` method to similarly clear the cell.  But then `highlight_char()` is called from `refresh_line_with_msg()` again, resetting the cell to the original value.

[...]

PR #729, opened 2023-08-29
No highligh_char on final refresh

gwenn:
Quick fix for #726


luketpeterson:
I have verified that this PR will fix the issue in https://github.com/kkawakam/rustyline/issues/726

Thank you!

PR #734, opened 2023-10-02
add Cmd::Repaint

sujiacong:
I need to rely on this refresh-only action when implementing a list function similar to bash complete. Do you think this PR can be merged into the main branch?

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -127,7 +127,7 @@ pub fn execute<H: Helper>(
         Cmd::Newline => {
             s.edit_insert('\n', 1)?;
         }
-        Cmd::Refresh => {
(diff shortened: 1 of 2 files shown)


gwenn:
`Repaint` seems to be the norm:
- https://docs.rs/reedline/latest/reedline/enum.ReedlineEvent.html#variant.Repaint
- https://docs.rs/termwiz/latest/termwiz/lineedit/enum.Action.html#variant.Repaint
- https://github.com/AmokHuginnsson/replxx/blob/1f149bfe20bf6e49c1afd4154eaf0032c8c2fda2/include/replxx.hxx#L207

gwenn:
Thanks.

PR #736, opened 2023-10-06
Make termios an optional dependency

### LINKED ISSUE #731: rustyline break some architecture on linux
The reason for this is that some of the terminal control constants in `termios` are wrong in some architectures, specifically at: https://github.com/dcuddeback/termios-rs/pull/37 but it looks like the termios authors don't show up much anymore, rustlyline Is it possible to fix the problem on your own code? The solution is simple: change all variables like `termios::VEOF` to `libc::VEOF`

gwenn:
Fix #731


gwenn:
@eatradish, @nospam3089
Could you please give it a try ?
Thanks.

eatradish:
Works for me

图片

gwenn:
> Works for me

Thanks.

PR #750, opened 2023-12-02
Support completion candidates that are shorter than the input

### LINKED ISSUE #748: CompletionType::List does not support completion candidates that are shorter than the input
Thank you for maintaining this crate — we use it in [Numbat](https://github.com/sharkdp/numbat) and it works great.

We want to implement a feature where a user could type something like `\hbar` in the REPL, press <kbd>Tab</kbd> and it would be replaced by `ℏ`. This seems to work fine when using `CompletionType::Circular`, but it does not work for `CompletionType::List`. Words will only be replaced if they are longer than the input. This is probably due to this check:

https://github.com/kkawakam/rustyline/blob/fafa0b9ba7cc6fbd21c055ea06ed42be6504ae62/src/lib.rs#L154

Would it be possible to support this use case?

Remark: that check above should probably use `lcp.unicode_width()` instead of `lcp.len()` (?).

gwenn:
Fix #748

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -151,7 +151,7 @@ fn complete_line<H: Helper>(
     } else if CompletionType::List == config.completion_type() {
         if let Some(lcp) = longest_common_prefix(&candidates) {
             // if we can extend the item, extend it
-            if lcp.len() > s.line.pos() - start {
+            if lcp.len() > s.line.pos() - start || candidates.len() == 1 {
                 completer.update(&mut s.line, start, lcp, &mut s.changes);
                 s.refresh_line()?;
             }
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -1,7 +1,7 @@
 use std::vec::IntoIter;
 
 use crate::completion::Completer;
-use crate::config::{Config, EditMode};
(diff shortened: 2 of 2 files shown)


sharkdp:
This works great — thank you very much!

I had to patch rustyline-derive in order for this branch to compile (not sure if this is the right fix though):
```patch
diff --git a/rustyline-derive/src/lib.rs b/rustyline-derive/src/lib.rs
index 5f57060..c25e616 100644
--- a/rustyline-derive/src/lib.rs
+++ b/rustyline-derive/src/lib.rs
@@ -126,8 +126,8 @@ pub fn highlighter_macro_derive(input: TokenStream) -> TokenStream {
                     ::rustyline::highlight::Highlighter::highlight_candidate(&self.#field_name_or_index, candidate, completion)
                 }
 
-                fn highlight_char(&self, line: &str, pos: usize) -> bool {
-                    ::rustyline::highlight::Highlighter::highlight_char(&self.#field_name_or_index, line, pos)
+                fn highlight_char(&self, line: &str, pos: usize, forced: bool) -> bool {
+                    ::rustyline::highlight::Highlighter::highlight_char(&self.#field_name_or_index, line, pos, forced)
                 }
             }
         }
```

PR #756, opened 2023-12-30
Add option to disable newline after every command

fcoury:
I am writing a neovim clone in Rust and the unconditional `writeln()?` on the `readline_with` method makes my whole screen scroll up every time the user enters a command.

I added a new `enable_newline` flag to the config, which defaults to true and created all the wrapper methods needed. Finally I changed the `readline_with` method to check the flag before issuing the `writeln` command.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -35,6 +35,8 @@ pub struct Config {
     check_cursor_position: bool,
     /// Bracketed paste on unix platform
     enable_bracketed_paste: bool,
+    /// Write new line after accepting prompt input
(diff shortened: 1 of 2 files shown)


gwenn:
See
https://www.javadoc.io/static/org.jline/jline/3.25.0/org/jline/reader/LineReader.Option.html#ERASE_LINE_ON_FINISH
So maybe we should use an enum instead of a boolean:
```rust
enum EndMode { // TODO find a better name
    Erase,
    NewLine,
    None,    
}
```

Issue #690, opened 2023-03-15
Multiple lines sent on pty input by `expect` are not seen by process


mmastrac:
When sending input via `expect`, `rustyline` doesn't appear to process anything but the first line.

`example.expect`

```
spawn cargo run --example numeric_input
send "1\n2\n"
expect "123"
```

Run `expect example.expect`

```
spawn cargo run --example numeric_input
1
2
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/examples/numeric_input`
> 1
Num: 1
> ^C✘-INT ~/Documents/github/rustyline [master|✔] 
```

Expected:

Output should be `Num 1` followed by `Num 2`.

Actual:

Output is just `Num 1`

Workaround:

It _does_ work without using `expect`, however:

```
$ echo -e "1\n2" | cargo run --example numeric_input
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/examples/numeric_input`
Num: 1
Num: 2
Error: Eof
```

gwenn:
I will try to reproduce.
Would you mind trying this setting: https://docs.rs/rustyline/latest/rustyline/config/enum.Behavior.html#variant.PreferTerm ?
```rust
    let config = Config::builder()
        .behavior(Behavior::PreferTerm)
        .build();
    let mut rl = DefaultEditor::with_config(config)?;
```

gwenn:
Ok, I can reproduce.
`Liner` seems OK.
But not `linenoise`, `replxx`, ...
I tried with `env_logger::init();` and `export RUST_LOG=rustyline=debug` but that didn't help.
It may be caused by the way we activate raw mode with `TCSADRAIN` [here](https://github.com/kkawakam/rustyline/blob/2c9e91154d3ba8afa0a5c9328f059c768ebe8ef9/src/tty/unix.rs#L1360) ?
I will try to investigate further...

gwenn:
Ok,
`rustyline` uses a [BufReader](https://doc.rust-lang.org/std/io/struct.BufReader.html).
And the whole input "1\n2\n" is read at once but only "1\n" is treated ("2\n" is discarded).
```diff
diff --git a/src/tty/unix.rs b/src/tty/unix.rs
index 23bfcce..70e02f4 100644
--- a/src/tty/unix.rs
+++ b/src/tty/unix.rs
@@ -144,6 +144,7 @@ impl Read for TtyIn {
                     return Err(error);
[...]

Issue #480, opened 2021-01-13
readline() can return multiple lines


rwestphal:
I've just found out that Editor::readline() can return multiple lines if one pastes a multi-line string from the clipboard into the terminal. Example:
```
root@debian:/mnt/renato/git/rust/rustyline# cargo run --example diy_hints
[snip]
This is a DIY hint hack of rustyline
> abc
def
ghi

input: abc
def
ghi

> 
```

Is this the expected behavior? If yes, is there a way I can change Editor::readline() to break the input into multiple lines (so that they can be parsed separately)?

gwenn:
You can either:
* [disable](https://github.com/kkawakam/rustyline/issues/202) bracketed paste mode. And you will received one line at a time.
* or split input yourself on `\n` (as far as I know, there will be distinction between a [continuation](https://docs.rs/rustyline/7.1.0/rustyline/enum.Cmd.html#variant.AcceptOrInsertLine) line and pasted text which contains newline(s)).

rwestphal:
> * [disable](https://github.com/kkawakam/rustyline/issues/202) bracketed paste mode. And you will receive one line at a time. You should be aware that `rustyline` is slow when pasted text is big.

@gwenn thanks for the quick response! That did the trick for me.

For future reference, here's how to disable the bracketed paste mode after the latest API changes:
```
rl.bind_sequence(KeyEvent(KeyCode::BracketedPasteStart, Modifiers::NONE), Cmd::Noop);
```

Closing...

PR #758, opened 2024-01-05
fix: bracketed paste multiline accept only one line

SalHe:
https://github.com/SalHe/rustyline/blob/e001cab7a04c36f2a113ef465325feb4c0d77f7b/src/lib.rs#L707

`RawReader` has been created and dropped when finished `readline_edit`.
`BufReader` has capacity 1024, so it might read more than `readline_edit` consumed (submitted when encouterred '\r').

For example, 
```
1
2
```
has been pasted. The buffer reads `"\033[220~1\r2\r\033[221~\000"`.
When accepted '\r', command has been submitted.

After this, we cannot retrieve rest of pasted command.
So '2\r\033[221~\000' has been ignored.

### DIFF
--- a/src/tty/unix.rs
+++ b/src/tty/unix.rs
@@ -233,7 +233,7 @@ impl PosixRawReader {
         pipe_reader: Option<PipeReader>,
     ) -> Self {
         Self {
-            tty_in: BufReader::with_capacity(1024, TtyIn { fd, sigwinch_pipe }),
+            tty_in: BufReader::with_capacity(1, TtyIn { fd, sigwinch_pipe }),
             timeout_ms: config.keyseq_timeout(),
             parser: Parser::new(),
             key_map,


gwenn:
See #690

gwenn:
Cannot reproduce locally:
```sh
% cargo run --example minimal
> line1
line2
line3
line4
line5
Line: line1
line2
line3
line4
line5
>
```
I tried with unix, mac, dos file content

SalHe:
> Cannot reproduce locally:
> 
> ```shell
> % cargo run --example minimal
> > line1
> line2
> line3
> line4
> line5
> Line: line1
> line2
> line3
> line4
> line5
> >
> ```
> 
> I tried with unix, mac, dos file content

Sorry I forgot to note that: `rl.bind_sequence(KeyEvent(KeyCode::BracketedPasteStart, Modifiers::NONE), Cmd::Noop);` should be configured.

```rust
use rustyline::{DefaultEditor, Result, Cmd, Modifiers, KeyCode, KeyEvent};

/// Minimal REPL
fn main() -> Result<()> {
    env_logger::init();
    let mut rl = DefaultEditor::new()?;
    rl.bind_sequence(KeyEvent(KeyCode::BracketedPasteStart, Modifiers::NONE), Cmd::Noop);
    loop {
        let line = rl.readline("> ")?; // read
        println!("Line: {line}"); // eval / print
    } // loop
}
```

In this case, pasted content should be accepted line by line. But now only the first line accepted.

SalHe:
It's a similar problem as #690 but without difference.
#690 solved lines sent to pty, but didn't solve `rl.bind_sequence(KeyEvent(KeyCode::BracketedPasteStart, Modifiers::NONE), Cmd::Noop);`. https://github.com/kkawakam/rustyline/issues/480#issuecomment-759759109

gwenn:
Sorry but your PR is fixing an edge case while impacting the common case.

gwenn:
@SalHe could you please try #761 ?
Or just [disable bracketed paster](https://docs.rs/rustyline/latest/rustyline/config/struct.Builder.html#method.bracketed_paste)

SalHe:
> @SalHe could you please try #761 ? Or just [disable bracketed paster](https://docs.rs/rustyline/latest/rustyline/config/struct.Builder.html#method.bracketed_paste)

Sorry for reply lately due to working stuff.
I have tried the latest `rustyline` with feature `buffer-redux` it does solve this problem. That's so great!
Thanks for your work.

Issue #348, opened 2020-03-08
Continuation prompt not used for multiline hints


thomcc:
You can see this in `cargo run --example example` after you get some multiline items in the history. It seems to be that the hint is just applied here directly, https://github.com/kkawakam/rustyline/blob/master/src/tty/unix.rs#L574-L580 (and basically the same for windows), instead of split up so that a continuation prompt can be applied.

Honestly it seems to me like it might be fine to just use spaces for the hint continuation prompt rather than further complicating prompt highlight code, otherwise it might be confusing... No strong feelings here though.

Issue #126, opened 2017-12-09
Add an async interface


Marwes:
Probably needs to spawn an OS thread and perform all line reading there with a channel (I don't think there is another way to do it anyway). I could (an likely will) implement this outside `rustyline` but if it seems useful enough to warrant inclusion I am happy to do a PR.

gwenn:
https://github.com/dpc/async-readline/blob/master/src/lib.rs#L165

rubdos:
I was thinking about tokio support, which basically means async support. I'm probably also trying to sketch something up.

gwenn:
https://docs.rs/tokio/0.1/tokio/io/struct.Stdin.html
> The handle implements the AsyncRead trait

gwenn:
@RadicalZephyr I don't know how to make sure `rustyline` and the other/main thread update the screen correctly ?

Ralith:
Very excited to see this receiving attention!

> I don't think Rustyline internals need to change to read from stdin asynchronously, I think it just needs to provide an async-compatible interface, which should probably be a futures [Stream](https://docs.rs/futures/0.1.25/futures/stream/trait.Stream.html) of `String`.

One reason to build async support into rustyline would be to allow operation without threads in cases where that is possible (e.g. stdin/out connected to a terminal or pipe on *nix). That's a lot of complexity for dubious benefit, though.

> I think for async use to be really nice, it would be good to have a new interface that allows output to be sent while the user is editing a line. I'm thinking of being able to create a remote "handle" to an `Editor` (kind of like the [`Handle`](https://docs.rs/tokio-core/0.1.17/tokio_core/reactor/struct.Handle.html) to a `Reactor` in the previous iteration of tokio. That handle would probably implement [`AsyncWrite`](https://docs.rs/tokio-io/0.1.11/tokio_io/trait.AsyncWrite.html).

Something like this sounds really great! Producing output without interfering with a prompt is exactly what I want for building e.g. an admin interface to a server. You might be able to get away with a synchronous `std::io::Write` interface, since nobody should be using this to dump large volumes of data anyway. Support for more fine-grained writes than complete lines will be useful for things like displaying live progress bars concurrent with user input.

remexre:
It'd also be nice to be able to use async for output from the log crate; a stretch-goal-feature for this could be a wrapper for Handle that impls [`Log`](https://docs.rs/log/0.4.6/log/trait.Log.html).

zyansheep:
[...]

PR #772, opened 2024-04-06
Add dynamic prompt support via ExternalPrinter

### LINKED ISSUE #417: Changing the prompt string dynamically

dmlary:
rustyline doesn't currently support changing the prompt while in the core readline loop.  There are a number of open PRs and issues for this functionality, but all of them appear to be stalled for more than a year.

Looking at #696 and 4ec26e8, the traditional appoach to this is to provide a reference to a trait object (`Prompt` or `ToString`), but with that appoach there's no way to cause the prompt to be redrawn for a change without user input.  This means for these appoaches the prompt could change without being displayed to the user.

There's an existing mechanism to allow another async task/thread to push input into the core readline loop, the `ExternalPrinter`.

In this commit, I expand `ExternalPrinter` to add `set_prompt()`.  With various plumbing, this function results in `wait_for_input` to return `Cmd::SetPrompt(String)`.

One of key change here is `State.prompt` changes from `&str` to `String`.  There is a performance hit here from the copy, but rustyline would need to prompt and receive input hundreds of times per second for the copy to have a noticable performance inpact.

Added examples/dynamic_prompt.rs to demonstrate the functionality.

Closes #417

Related #208, #372, #369, #417, #598, #696

### DIFF
--- a/examples/dynamic_prompt.rs
+++ b/examples/dynamic_prompt.rs
@@ -1,0 +1,25 @@
+use std::thread;
+use std::time::Duration;
+
+use rustyline::{DefaultEditor, ExternalPrinter, Result};
+
+fn main() -> Result<()> {
+    let mut rl = DefaultEditor::new()?;
+    let mut printer = rl.create_external_printer()?;
+    thread::spawn(move || {
+        let mut i = 0usize;
+        loop {
+            printer
+                .set_prompt(format!("prompt {:02}>", i))
+                .expect("set prompt successfully");
+            thread::sleep(Duration::from_secs(1));
+            i += 1;
+        }
+    });
+
+    loop {
+        let line = rl.readline("> ")?;
+        rl.add_history_entry(line.as_str())?;
+        println!("Line: {line}");
+    }
+}
--- a/src/command.rs
+++ b/src/command.rs
@@ -16,7 +16,7 @@ pub enum Status {
 
 pub fn execute<H: Helper>(
     cmd: Cmd,
-    s: &mut State<'_, '_, H>,
+    s: &mut State<'_, H>,
     input_state: &InputState,
     kill_ring: &mut KillRing,
     config: &Config,
@@ -228,6 +228,10 @@ pub fn execute<H: Helper>(
             // the input
             s.move_cursor_to_end()?;
             return Err(error::ReadlineError::Interrupted);
+        }
+        Cmd::SetPrompt(prompt) => {
+            s.set_prompt(prompt);
+            s.refresh_line()?;
         }
         _ => {
             // Ignore the character typed.
--- a/src/edit.rs
+++ b/src/edit.rs
@@ -23,9 +23,9 @@ use crate::KillRing;
 
 /// Represent the state during line editing.
 /// Implement rendering.
-pub struct State<'out, 'prompt, H: Helper> {
+pub struct State<'out, H: Helper> {
     pub out: &'out mut <Terminal as Term>::Writer,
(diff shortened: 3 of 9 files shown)


dmlary:
Note, this is only the unix implementation.  I do not have access to a windows development environment to add the implementation for that side.  It's not a complicated change, but for now it's marked as `unimplemented!()`.

gwenn:
See https://github.com/kkawakam/rustyline/issues/126#issuecomment-1887685910
> removing / replacing [ExternalPrinter](https://docs.rs/rustyline/latest/rustyline/trait.ExternalPrinter.html) by the clever / simple [linenoise](https://github.com/antirez/linenoise#asyncrhronous-api) solution is appealing.

And another issue related to the prompt: #348
And a more fundamental issue: https://github.com/kkawakam/rustyline/blob/master/Incremental.md
> Ideally, we would like to redraw only impacted row(s) / cell(s).

So, for me, when *only* the prompt (`Cmd::SetPrompt`) is changed, we should not have to redraw the whole line (` s.refresh_line()`) or at least we should **not** have to invoke `Hinter::hint()` / `Highlighter::highlight_char` / `Highlighter::highlight` (only `Highlighter::highlight_prompt`).

I apologize for the stalled situation induced by some bad design choices.

dmlary:
TL;DR: Pushing a change to reduce processing when prompt changes.  Linenoise approach is out-of-scope for this PR, but I am looking at how to do something similar.

### Reduce computation for `Cmd::SetPrompt`
> removing / replacing [ExternalPrinter](https://docs.rs/rustyline/latest/rustyline/trait.ExternalPrinter.html) by the clever / simple [linenoise](https://github.com/antirez/linenoise#asyncrhronous-api) solution is appealing.

I dug into the linenoise code to look at this note.  Linenoise does not directly support changing the prompt.  Although the prompt is stored in the `struct linenoiseState`, the only way to get the prompt to be redrawn is by calling `linenoiseShow()` at minimum.  This causes a full redraw of the prompt.  Which leads to...

> [...] when only the prompt (Cmd::SetPrompt) is changed, we should not have to redraw the whole line [...]

I don't think this is advisable when you consider that the length of the prompt may change.  In that case you'd have to redraw the rest of the line as all of the characters would need to be shifted.   You can absolutely implement this, but it's added complexity for sub-millisecond performance gain on user keyboard input.

> when only the prompt (Cmd::SetPrompt) is changed, [...] or at least we should not have to invoke Hinter::hint() / Highlighter::highlight_char / Highlighter::highlight (only Highlighter::highlight_prompt)

This makes sense as the `hint()` function may do blocking things like checking networked filesystems.  I made this change by adding `Refresher::refresh_prompt()` as a cut down version of `refresh_line()`.  I could instead make the call to `State::refresh()` in `State::set_prompt()`, but it looked like you wanted to maintain a clear separation.

### Remove `ExternalPrinter`
I agree with you that the linenoise approach where the caller only calls into rustyline when there is input to be processed is a much better approach.  I've already hit some additional limitations with `ExternalPrinter` in my project, so I'll see if there's a simple way to implement that for rustyline.  No guarantees on my timeline here.

That said, I think adding that interface is out of scope for this PR.  It's an architectural shift in rustyline, and not directly related to adding dynamic prompt support.

xeruf:
Hey, would love to see progress here!

PR #776, opened 2024-05-01
avoid panic: `entry[pos..]` => `entry.get(pos..)`

lengyijun:


gwenn:
I may be wrong but:
We have:
```rust
        if line.is_empty() || pos < line.len() {
            return None;
        }
```
So `pos == line.len() && pos > 0`.
So I guess that the panic can happen only if `history` returns an entry that **doesn't** start with `line`.
Which seems fine to me.

PR #786, opened 2024-07-28
Read cursor position with cluttered input

gwenn:
Try to read cursor position even when stdin contains / receives user input. Currently, we fail so that we don't discard any user input. But we should be able to fill the `BufReader` instead...


gwenn:
Cannot reproduce

PR #793, opened 2024-08-12
User-specified continuation prompt

zao111222333:
I believe the user-specified continuation prompt is a crucial feature, especially in scenarios where indentation is important (such as Python shell). I have implemented this feature by replacing the '\n' character in multi-line input with `\n`+`continuation_prompt` in display phase (in `Renderer::refresh_line`). This approach can cause a misalignment between the displayed cursor position and the actual editing position. To address this, I used `State::continuation_prompt_move_cursor` to correct these misalignments.

Additionally, I have added an example. You can run it with the following command:

``` shell
cargo run --release --example continuation_prompt --features="custom-bindings derive"
```
And the demo results:
image

However, the current implementation only meets the minimum use case (i.e., continuation prompt works when pasting multi-line text into rustyline), but I have still encountered some bugs:

+ Cursor misalignment when inserting/deleting text.
+ Cursor misalignment when moving the cursor to the top or bottom of the input.
+ Cursor misalignment when using `Cmd::Newline`.

Since I'm not very familiar with rustyline, I want to share my current implementation and seek help and suggestions here. Thank you!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,6 +83,9 @@ required-features = ["custom-bindings", "derive"]
 name = "input_multiline"
 required-features = ["custom-bindings", "derive"]
 [[example]]
+name = "continuation_prompt"
+required-features = ["custom-bindings", "derive"]
+[[example]]
 name = "input_validation"
 required-features = ["derive"]
 [[example]]
--- a/examples/continuation_prompt.rs
+++ b/examples/continuation_prompt.rs
@@ -1,0 +1,35 @@
+use rustyline::validate::MatchingBracketValidator;
+use rustyline::{Cmd, Editor, EventHandler, KeyCode, KeyEvent, Modifiers, Result, highlight::Highlighter};
+use rustyline::{Completer, Helper, Hinter, Validator};
+use std::borrow::Cow::{self, Borrowed};
+#[derive(Completer, Helper, Hinter, Validator)]
+struct InputValidator {
+    #[rustyline(Validator)]
+    brackets: MatchingBracketValidator,
+}
+
+impl Highlighter for InputValidator {
+    fn continuation_prompt<'p, 'b>(
+        &self,
+        prompt: &'p str,
+        default: bool,
+    ) -> Option<Cow<'b, str>> {
+        Some(Borrowed(".... "))
+    }
+}
+
+fn main() -> Result<()> {
+    let h = InputValidator {
+        brackets: MatchingBracketValidator::new(),
+    };
+    let mut rl = Editor::new()?;
+    rl.set_helper(Some(h));
+    rl.bind_sequence(
+        KeyEvent(KeyCode::Char('s'), Modifiers::CTRL),
+        EventHandler::Simple(Cmd::Newline),
+    );
+    let input = rl.readline(">>>> ")?;
+    println!("Input:\n{input}");
+
+    Ok(())
+}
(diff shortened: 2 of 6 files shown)


gwenn:
See #372
And https://github.com/gwenn/rustyline-notes/blob/master/src/design.md

zao111222333:
> See #372 And https://github.com/gwenn/rustyline-notes/blob/master/src/design.md

Hi, I agree with "split the input by line" mentioned in [design.md](https://github.com/gwenn/rustyline-notes/blob/master/src/design.md), the implementation here is essentially split the input in renderer (soft wrapped), and then re-align cursor.

And I'd like to contribute for continuation prompt if the new `Vec<(Range,Style)>` API is finished.

gwenn:
We might be able to fix the `Highlighter` API like [isocline](https://github.com/daanx/isocline/blob/c9310ae58941559d761fe5d2dd2713d245f18da6/include/isocline.h#L283-L294).
And also check how `isocline` handles continuation prompt:
- [ic_get_continuation_prompt_marker](https://github.com/daanx/isocline/blob/c9310ae58941559d761fe5d2dd2713d245f18da6/include/isocline.h#L331)
- [ic_enable_multiline_indent](https://github.com/daanx/isocline/blob/c9310ae58941559d761fe5d2dd2713d245f18da6/include/isocline.h#L362)

gwenn:
Also I don't want you to waste your time.
So maybe you should check if some [alternatives](https://github.com/gwenn/rustyline-notes/blob/master/src/alternatives.md) already supports continuation prompt.
For example: https://github.com/nushell/reedline/issues/112

zao111222333:
> Also I don't want you to waste your time. So maybe you should check if some [alternatives](https://github.com/gwenn/rustyline-notes/blob/master/src/alternatives.md) already supports continuation prompt. For example: [nushell/reedline#112](https://github.com/nushell/reedline/issues/112)

Thank you a lot for your time and information, I will have a try for those alternatives. 
Also I will still try to implement this feature for rustyline.

gwenn:
For the highlighting part, we might use this library:
https://crates.io/crates/ansi-str
> A library which provides a set of methods to work with ANSI strings
> It's an agnostic library in regard to different color libraries.

This method seems to do the job:
https://docs.rs/ansi-str/latest/ansi_str/fn.get_blocks.html

zao111222333:
> Just in case if you are still interested,
> 
> Could you please have a look at:
> 
> https://github.com/kkawakam/rustyline/pull/795/commits/6b34d32cb43561856d538bc729ef8cfd0d7de5aa
> 
> ?
> 
> I guess it should be feasible like the [Split](https://doc.rust-lang.org/std/str/struct.Split.html) iterator but I am not sure.

Thank you for informing, I will try it :)

gwenn:
> > Just in case if you are still interested,
> > Could you please have a look at:
> > [6b34d32](https://github.com/kkawakam/rustyline/commit/6b34d32cb43561856d538bc729ef8cfd0d7de5aa)
> > ?
> > I guess it should be feasible like the [Split](https://doc.rust-lang.org/std/str/struct.Split.html) iterator but I am not sure.
> 
> Thank you for informing, I will try it :)

Sorry I mean help me to fix it if you can...
(you cannot give it a try because it doesn't type-check)

zao111222333:
> > > Just in case if you are still interested,
> > > Could you please have a look at:
> > > [6b34d32](https://github.com/kkawakam/rustyline/commit/6b34d32cb43561856d538bc729ef8cfd0d7de5aa)
> > > ?
> > > I guess it should be feasible like the [Split](https://doc.rust-lang.org/std/str/struct.Split.html) iterator but I am not sure.
> > 
> > 
> > Thank you for informing, I will try it :)
> 
> Sorry I mean help me to fix it if you can... (you cannot give it a try because it doesn't type-check)

Oh I see, sure I can help to try how to fix it (might be this weekend). Thanks!

PR #795, opened 2024-08-15
Refactor highlighting

gwenn:
Introduce `Style` and `StyledBlock` traits
And their `ansi-str` implementations

- [ ] Make `ansi-str` optional
- [ ] Fix `Highlighter` trait
- [ ] Fix rendering accordingly
- [ ] Check that new API works with an hardcoded continuation prompt.

See #793, #372
See https://github.com/gwenn/rustyline-notes/blob/master/src/design.md

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,9 @@ maintenance = { status = "actively-developed" }
 members = ["rustyline-derive"]
 
 [dependencies]
+# For highlighting
+# TODO make it optional
+ansi-str = { version = "0.8.0", optional = false }
 bitflags = "2.6"
 cfg-if = "1.0"
 # For file completion
--- a/src/highlight.rs
+++ b/src/highlight.rs
@@ -3,6 +3,46 @@
 use crate::config::CompletionType;
 use std::borrow::Cow::{self, Borrowed, Owned};
 use std::cell::Cell;
+use std::fmt::Display;
+
+/// ANSI style
+pub trait Style {
+    /// Produce a ansi sequences which sets the graphic mode
+    fn start(&self) -> impl Display;
+    /// Produce a ansi sequences which ends the graphic mode
+    fn end(&self) -> impl Display;
+}
+
+/// Styled text
+pub trait StyledBlock {
+    /// Style impl
+    type Style: Style;
+    /// Raw text to be styled
+    fn text(&self) -> &str;
+    /// `Style` to be applied on `text`
+    fn style(&self) -> &Self::Style;
+}
+
+impl Style for ansi_str::Style {
+    fn start(&self) -> impl Display {
+        self.start()
+    }
+
+    fn end(&self) -> impl Display {
+        self.end()
+    }
+}
+impl StyledBlock for ansi_str::AnsiBlock<'_> {
+    type Style = ansi_str::Style;
+
+    fn text(&self) -> &str {
+        self.text()
+    }
+
+    fn style(&self) -> &Self::Style {
+        self.style()
+    }
+}
 
 /// Syntax highlighter with [ANSI color](https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_(Select_Graphic_Rendition)_parameters).
 /// Rustyline will try to handle escape sequence for ANSI color on windows
@@ -277,5 +317,16 @@ mod tests {
         use super::is_open_bracket;
         assert!(is_open_bracket(b'('));
         assert!(is_close_bracket(b')'));
+    }
+
+    #[test]
+    pub fn styled_text() {
+        use ansi_str::get_blocks;
+
+        let mut blocks = get_blocks("\x1b[1;32mHello \x1b[3mworld\x1b[23m!\x1b[0m");
+        assert_eq!(blocks.next(), get_blocks("\x1b[1;32mHello ").next());
+        assert_eq!(blocks.next(), get_blocks("\x1b[1;32m\x1b[3mworld").next());
+        assert_eq!(blocks.next(), get_blocks("\x1b[1;32m!").next());
+        assert!(blocks.next().is_none())
     }
 }


zao111222333:
It seems that we can change the `ansi_str::get_blocks` form `fn get_blocks(text: &str)` to `fn get_blocks<'s>(text: Cow<'s, str>)`. And then all those error here [6b34d32](https://github.com/kkawakam/rustyline/pull/795/commits/6b34d32cb43561856d538bc729ef8cfd0d7de5aa) will be solved.
(In this case we have to create a new crate rather than use `ansi_str` since I don't know the origin author's opinion with this change). 
Shall I have a try on this approach and make a commit?

zao111222333:
I find that once we define 
``` rust
fn highlight_line<'l>(&self, line: &'l str, pos: usize) -> impl Iterator<Item = impl 'l+StyledBlock>
```

The new `highlight_line` function signature conflicts to the `Option<&dyn Highligher>` due to https://doc.rust-lang.org/reference/items/traits.html#object-safety . So use generic `<H: Highligher>` + `Option<&H>` to replace `Option<&dyn Highligher>`

And I have maked a pull request, see https://github.com/kkawakam/rustyline/pull/799

gwenn:
Maybe we should not try to implement `highlight_line` for `ansi_str`.

zao111222333:
Looks good! And how do you think `Vec<(Self::Style, &'l str)>` or `impl Iterator<(Self::Style, &'l str)>`?
The cons and pros of `impl Iterator<(Self::Style, &'l str)>`:
pros:
+ More flexible, a superset of `Vec` since `Vec::<Foo>::into_iter` will return `impl Iterator<Foo>`
+ Can be a lazy iterators (e.g., `Map`, `FilterMap`, etc.), since `let v: Vec<_> = iter().map(...).collect()` require a duplicate `collect` function with **O(n)**
cons:
+ Not straightforward for beignner (compare to `Vec`)
+ Conflicts to the `Option<&dyn Highligher>` due to https://doc.rust-lang.org/reference/items/traits.html#object-safety, but we can use `<H: Highligher>` + `Option<&H>` to overcome this issue

gwenn:
Alternatives use `Vec`:
https://docs.rs/syntect/5.2.0/syntect/easy/struct.HighlightLines.html#method.highlight
https://docs.rs/reedline/latest/reedline/trait.Highlighter.html#tymethod.highlight + https://docs.rs/reedline/latest/reedline/struct.StyledText.html
https://docs.rs/termwiz/latest/termwiz/lineedit/trait.LineEditorHost.html#method.highlight_line

gwenn:
`Highlighter` implementations are becoming messy.
See 9e5c0e1
Maybe we should remove the `type Style` associated type and support only `()` or `anstyle::Style` ?

zao111222333:
Thanks a lot for your hard working, I believe `impl Iterator<Item = impl 'l + StyledBlock>` is the best function signature. And what can I do for this feature?

gwenn:
> And what can I do for this feature?

See above:
[ ] Fix rendering accordingly
[ ] Check that new API works with an hardcoded continuation prompt.

zao111222333:
I seem to find a decent way to finish the task "Check that new API works with an hardcoded continuation prompt",
with the pull request https://github.com/gwenn/rustyline/pull/3

zao111222333:
I am sorry to make so many changes in this commit, but I have to explain the reason:

1. I impl `StyledBlock` for `(syntect::highlighting::Style,&str)`.
   `syntect` is a syntax highlighting library using Sublime Text's grammars, with this crate we can use Sublime ecosystem for almost all languages' highlighting.

2. I remove the feature gating logics controlled by `anstyle` or `ansi_str`.
   I believe the `split-highlight` should be an all-in-one general gate control feature, and  those ansi-implemented crates could exist together.
   Users can chose the ansi-implemented crate (`anstyle::Style`/`ansi_str::Style`/`syntect::highlighting::Style`) they prefer to.
   And I re-expose `anstyle`, `ansi_str`, `syntect` in `crate::highlight`.

3. To verify the continuation prompt and syntax highlighting, I add an python highlight example:

   highhlight_python

Expecting your suggestions & feedback

zao111222333:
After weeks of struggling, here is my solution, pls have a look at https://github.com/gwenn/rustyline/pull/3

zao111222333:
Let me briefly summarize my commits
1. Use [`DisplayOnce`](https://github.com/zao111222333/rustyline/blob/496d2767ef5168d1629ae7f7cc712f8e6d8000cf/src/highlight.rs#L17-L41) to have a consistent & back-compatible highlight output type
``` rust
fn highlight<'b, 's: 'b, 'l: 'b>(
    &'s mut self,
    line: &'l str,
    pos: usize,
) -> impl 'b + DisplayOnce;
```
the back-compatiblity is that, the old highligh function even needn't to change signature, since we will auto-impl `DisplayOnce` for all types that impl `Display`. for [example](https://github.com/zao111222333/rustyline/blob/496d2767ef5168d1629ae7f7cc712f8e6d8000cf/examples/read_password.rs#L12-L16)
``` rust
fn highlight<'b, 's: 'b, 'l: 'b>(
    &'s mut self,
    line: &'l str,
    pos: usize,
) -> std::borrow::Cow<'b, str>
```
And it also supports **split-highlight**, since we impl `DisplayOnce` for [`StyledBlocks`](https://github.com/zao111222333/rustyline/blob/496d2767ef5168d1629ae7f7cc712f8e6d8000cf/src/highlight.rs#L69-L96), a wrapper of `StyledBlock` iterator.
```rust
pub struct StyledBlocks<'l, B, I>
where
    B: 'l + StyledBlock,
    I: Iterator<Item = B>,
{
    iter: I,
    _marker: PhantomData<&'l ()>,
}
```
2. Reduce tokenizer/parser overhead
I don't get the point why we need `helper: Option<H>` in `Editor`, and I change it to `helper: H`. Then we can avoid lifetime issue when using `&mut self` in `highlight`, `validate`, `complete`, ...
as you have mentioned at
https://github.com/gwenn/rustyline/blob/highlight/src/completion.rs#L93
https://github.com/gwenn/rustyline/blob/highlight/src/lib.rs#L543
Now the `Helper` becomes https://github.com/zao111222333/rustyline/blob/496d2767ef5168d1629ae7f7cc712f8e6d8000cf/src/lib.rs#L548-L584
User can put the tokenizer/parser at `update_after_edit` to update context, then just use the context in `highlight`, `validate`, `complete`, ...
```rust
fn update_after_edit(&mut self, line: &str, pos: usize, forced_refresh: bool)
```
3. hardcoded continuation prompt
User can implement the continuation prompt in `highlight` function to render continuation prompt. And then set the continuation prompt width for shifting cursor when editing, by [`Helper::continuation_prompt_width`](https://github.com/zao111222333/rustyline/blob/496d2767ef5168d1629ae7f7cc712f8e6d8000cf/src/lib.rs#L578-L583)

Issue #629, opened 2022-05-01
External printer sometimes omits messages until more input is given


JanBeh:
I'm using FreeBSD and xterm and tried the new [external printer](https://github.com/kkawakam/rustyline/blob/280e71191fef234c698fe1d94dedc270a6f37c33/examples/external_print.rs) feature.

When I type in text, the program and/or terminal sometimes gets into a state where no more messages are being emitted. I used [this example code](https://github.com/kkawakam/rustyline/blob/280e71191fef234c698fe1d94dedc270a6f37c33/examples/external_print.rs), with slightly adjusted timings:

```diff
-            let wait_ms = rng.gen_range(1000..10000);
+            let wait_ms = rng.gen_range(500..1000);
```

This is how my terminal looks like after I type in some text when I'm unlucky:

```text
% cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/readnwrite`
External message #0
External message #1
External message #2
External message #3
> ghdflghdfjkhd
```

At this point, no more output will be created, even if I wait for a long time. I then typed a few `x` characters, and this happened:

```text
% cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/readnwrite`
External message #0
External message #1
External message #2
External message #3
External message #32
External message #33
External message #34
External message #35
External message #36
External message #37
External message #38
> ghdflghdfjkhdfxxxxxx
Error: Interrupted
```

Note that also an `f` appeared after `ghdflghdfjkhd` and before the `xxxxxx`. That `f` must have been stuck somewhere because I didn't enter it after the state appeared to be blocked.

Messages #4 through #31 got lost in this example.

My setup:

```text
% echo $TERM
xterm
% xterm -v
XTerm(372)
```

gwenn:
Sorry but I cannot reproduce on linux/macosx:
```sh
$ uname -a
Linux archlinux 5.17.5-arch1-1 #1 SMP PREEMPT Wed, 27 Apr 2022 20:56:11 +0000 x86_64 GNU/Linux
```
```sh
% uname -a
Darwin Gwenaels-MacBook-Air.local 20.6.0 Darwin Kernel Version 20.6.0: Tue Feb 22 21:10:41 PST 2022; root:xnu-7195.141.26~1/RELEASE_X86_64 x86_64
% echo $TERM
xterm-256color
```
I will try to investigate on a FreeBSD VM...
[...]

PR #798, opened 2024-08-23
Implement ExternalPrinterWriter

### LINKED ISSUE #797: Implement Write for external printer
So it can be used as pipe for log output, as I implemented here: https://github.com/Dhole/rustylinez/pull/2/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759

xeruf:
To enable duplex output of log messages

Fixes #797

### DIFF
--- a/examples/external_print.rs
+++ b/examples/external_print.rs
@@ -1,5 +1,11 @@
+extern crate env_logger;
+extern crate log;
+
+use std::io::Write;
 use std::thread;
 use std::time::Duration;
+use env_logger::Target;
+use log::LevelFilter;
 
 use rand::{thread_rng, Rng};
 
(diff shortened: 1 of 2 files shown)


xeruf:
Only Problem: When `rustyline` logs at debug level, the prompt somehow does not work. Fixed in my project by `.filter(Some("rustyline"), LevelFilter::Warn)`

gwenn:
I am afraid that on incomplete / partial write, `rustyline` is going to append newline (\n) where they are not expected.
See https://github.com/kkawakam/rustyline/blob/e4eca97cbbff139278c153e40aa285920d2637dd/src/edit.rs#L324-L326

xeruf:
That is why the method was called `println` in the rustylinez fork, but that is not the main problem even though it is not quite ideal.
The readline function completely hangs when at debug level for rustyline, probably they somehow interfere with each other.

gwenn:
Indeed, it should be named `println`.
But initially, there was no newline appended automatically until:
https://github.com/kkawakam/rustyline/issues/629
https://github.com/kkawakam/rustyline/pull/627

xeruf:
anyways, I am using this modification despite the issues mentioned in this thread, so feel free to build on it or not, as unfortunately I do not have the resources to polish it properly
just wanted to share my work upstream :)

Issue #607, opened 2022-03-03
Add support for cancelling input with Escape


PaulJuliusMartinez:
I think in certain contexts it make sense for `Escape` to cancel entering an input, similar to Ctrl-C or Ctrl-D. When entering a `:` command in vim, for example, hitting escape will return you to normal mode.

gwenn:
But rustyline has only normal and insert mode, no command line mode.
And on unix, a single escape is always ambiguous: escape sequence versus single escape depending on how fast you type.
Have you try to [rebind](https://docs.rs/rustyline/latest/rustyline/struct.Editor.html#method.bind_sequence) this key ?

schungx:
Rebinding the Escape key works fine (you essentially make it the same as <kbd>Ctrl</kbd>-<kbd>U</kbd>.

gwenn:
But with Windows Terminal, you may have the same issue as on unix if we activate `ENABLE_VIRTUAL_TERMINAL_INPUT`...

schungx:
> But with Windows Terminal, you may have the same issue as on unix if we activate `ENABLE_VIRTUAL_TERMINAL_INPUT`...

Not really.  I have mapped <kbd>Esc</kbd> to clear-line like this and it works just fine:

```rust
    // On Windows, Esc clears the input buffer
    #[cfg(target_family = "windows")]
    rl.bind_sequence(
        Event::KeySeq(smallvec![KeyEvent(KeyCode::Esc, Modifiers::empty())]),
        EventHandler::Simple(Cmd::Kill(Movement::WholeBuffer)),
    );
```

PaulJuliusMartinez:
> But rustyline has only normal and insert mode, no command line mode.

I wasn't referring to vim-mode in Rustyline, just using vim as an example of a program that allows using Escape to cancel entering input in readline-like contexts. You can also use Escape to cancel entering a search pattern after pressing `/` in vim.

I understand that escapes are fundamentally ambiguous, but there are workarounds for that -- lots of programs allow configuring a timeout where, if no data is supplied after reading an Escape byte, it'll register as an Escape press. And those programs also often support setting that timeout to 0. I think the rough assumption is that you'll basically never read half an escape sequence, so if you ask to read 64 bytes, but only get a single escape byte back, it's pretty likely that's an actual Escape press.


I tried binding a sequence, but it didn't seem to work:
```rust
rustyline_editor.bind_sequence(
    KeyEvent::new('\x1B', Modifiers::empty()),
[...]

PR #802, opened 2024-09-08
Don't call poll when keyseq_timeout is set to 0 to avoid broken poll implementation for devices on MacOS.

PaulJuliusMartinez:
The poll system call is [broken on MacOS](https://nathancraddock.com/blog/macos-dev-tty-polling/) and doesn't work properly for devices, such as `/dev/tty`.

As a result, reading single escapes is broken when using `Behavior::PreferTerm`, which causes `rustyline` to read from `/dev/tty`. After reading a single escape character, the subsequent call to `poll` will always return `Ok(1)`, so `rustyline` will try to parse an escape sequence.

This PR makes a small change to at least improve the behavior when `keyseq_timeout` is set to `Some(0)`. In this case, we bypass the call to `poll` altogether, and instead check whether there is any buffered input to determine whether a single escape keypress has been entered or if an escape sequence has been entered, which is a pretty effective heuristic.

I don't know if there's any good workaround for the broken poll implementation when `keyseq_timeout` is non-zero, but this is at least an improvement.

This should resolve https://github.com/kkawakam/rustyline/issues/607.

### DIFF
--- a/src/tty/unix.rs
+++ b/src/tty/unix.rs
@@ -799,16 +799,26 @@ impl RawReader for PosixRawReader {
             } else {
                 self.timeout_ms
             };
-            match self.poll(timeout_ms) {
-                Ok(0) => {
-                    // single escape
-                }
-                Ok(_) => {
-                    // escape sequence
+            if timeout_ms == PollTimeout::ZERO {
+                // Don't bother polling at all if [timeout_ms] is 0. (Also because
+                // calling [poll] on /dev/tty on MacOS is broken.) If there's any
+                // buffered input we'll treat it as an escape sequence, and otherwise
+                // we'll assume it's a single escape.
+                if !self.tty_in.buffer().is_empty() {
                     key = self.escape_sequence()?
                 }
-                // Err(ref e) if e.kind() == ErrorKind::Interrupted => continue,
-                Err(e) => return Err(e),
+            } else {
+                match self.poll(timeout_ms) {
+                    Ok(0) => {
+                        // single escape
+                    }
+                    Ok(_) => {
+                        // escape sequence
+                        key = self.escape_sequence()?
+                    }
+                    // Err(ref e) if e.kind() == ErrorKind::Interrupted => continue,
+                    Err(e) => return Err(e),
+                }
             }
         }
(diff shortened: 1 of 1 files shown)


gwenn:
I manage to reproduce the issue.
Only with this patch to debug:
```diff
diff --git a/src/tty/unix.rs b/src/tty/unix.rs
index 08d7548..264fc3a 100644
--- a/src/tty/unix.rs
+++ b/src/tty/unix.rs
@@ -707,8 +707,10 @@ impl PosixRawReader {
         if n > 0 {
             return Ok(n as i32);
         }
+        debug!(target: "rustyline", "poll with: {:?}", timeout_ms);
         let mut fds = [poll::PollFd::new(self.as_fd(), PollFlags::POLLIN)];
         let r = poll::poll(&mut fds, timeout_ms);
+        debug!(target: "rustyline", "poll returns: {:?}", r);
         match r {
             Ok(n) => Ok(n),
             Err(Errno::EINTR) => {
```

And this example (**timeout <> 0** + `Behavior::PreferTerm `):
```rust
use rustyline::{Config, DefaultEditor, Result};

fn main() -> Result<()> {
    env_logger::init();
        .behavior(rustyline::Behavior::PreferTerm)
        .keyseq_timeout(Some(50))
        .build();
    let mut rl = DefaultEditor::with_config(config)?;

    let _ = rl.readline("> ")?;
    Ok(())
}
```
```sh
RUST_LOG=rustyline=debug cargo run --example issue 2> debug.log
```
gives
```
[DEBUG rustyline] poll with: PollTimeout(50)
[DEBUG rustyline] poll returns: Ok(1)
```
on MacOS as you say.
But we **cannot** rely on timeout.

No issue with `Behavior::Stdio`:
```
[DEBUG rustyline] poll with: PollTimeout(50)
[DEBUG rustyline] poll returns: Ok(0)
[DEBUG rustyline] c: '\u{1b}' => key: KeyEvent(Esc, Modifiers(0x0))
[DEBUG rustyline] Emacs command: Abort
```

(I am going to check on Linux)

gwenn:
See #803

gwenn:
Would you mind reviewing / testing #803 ?

PR #808, opened 2024-10-04
Implement `Candidate` for `Arc<str>`

fsktom:
Should've already done it in #721


gwenn:
See #810

fsktom:
Thanks @gwenn ! 👍🏿

PR #809, opened 2024-10-05
mutable `Helper` in functions

zao111222333:
See https://github.com/gwenn/rustyline/pull/4
And I add a new `example/bracket_parser.rs` to demo `Parser` trait.
bracket_parser

btw, rightnow I have no idea how to represent `change: InputEdit`.


gwenn:
Thank you for all your proposals.
But please only one feature at a time: either mut Helper **or** Parser but not both in the same PR.

src/lib.rs
```diff
@@ -482,7 +492,7 @@ fn apply_backspace_direct(input: &str) -> String {
 fn readline_direct(
     mut reader: impl BufRead,
     mut writer: impl Write,
-    validator: &Option<impl Validator>,
+    mut validator: Option<&mut impl Validator>,
```
gwenn, line 495: not sure why you need the `mut validator`

src/edit.rs
```diff
@@ -256,29 +296,36 @@ impl<'out, 'prompt, H: Helper> State<'out, 'prompt, H> {
     }
 }
 
-impl<'out, 'prompt, H: Helper> Invoke for State<'out, 'prompt, H> {
-    fn input(&self) -> &str {
-        self.line.as_str()
-    }
-}
+// We only use `self.line` to create `impl Invoke`.
```
gwenn, line 299: See https://github.com/kkawakam/rustyline/blob/ceafb7f8057974201d0e76ac9fc7aaac3998a048/src/validate.rs#L44-L47

rustyline-derive/src/lib.rs
```diff
@@ -102,32 +102,42 @@ pub fn highlighter_macro_derive(input: TokenStream) -> TokenStream {
         quote! {
             #[automatically_derived]
             impl #impl_generics ::rustyline::highlight::Highlighter for #name #ty_generics #where_clause {
-                fn highlight<'l>(&self, line: &'l str, pos: usize) -> ::std::borrow::Cow<'l, str> {
-                    ::rustyline::highlight::Highlighter::highlight(&self.#field_name_or_index, line, pos)
+                #[cfg(any(not(feature = "split-highlight"), feature = "ansi-str"))]
```
gwenn, line 105: No `split-highlight` in this PR please

gwenn (changes requested):
Remove all `Parser` / `split-highlight` related code

gwenn:
See https://github.com/kkawakam/rustyline/blob/ceafb7f8057974201d0e76ac9fc7aaac3998a048/src/undo.rs#L9 for a possible `InputEdit` impl (except for `Begin` / `End`)

rustyline-derive/src/lib.rs
zao111222333: I am sorry for that, I directly copy `rustyline-derive/src/lib.rs` for the old version. I only copy this single file and I will remove those :)

gwenn: No worry !

src/lib.rs
zao111222333: The `match` block requires the input `validator` is mutable. And when we call it, use `readline_direct(io::stdin().lock(), io::stderr(), self.helper.as_mut())`, it is almost same as before
https://github.com/kkawakam/rustyline/blob/77e7d3d772c2db44307a559dfeb0623e44d25165/src/lib.rs#L519-L523

gwenn: Ok

zao111222333:
> But please only one feature at a time: either mut Helper or Parser but not both in the same PR.

Understand, and since the `Parser::parse` requires `&mut self`, I think I can firstly fix all issue (like create `ValidationContext` from `State`) for mut Helper. Merge it firstly.
And then create another PR for `Parser` (as well as `InputEdit`)

zao111222333:
Here is the update under your suggestions.
I think the `Invoke` should impl for `LineBuffer` and `String`, rather than `State` and `&str`.
The `&str` could be replaced to `String` since itself will be changed in `invoke`.
And `State` could be replaced to `LineBuffer` since `Invoke` seems only require `LineBuffer`.
How do you think that?

I haven't fixed `fn invoke(&mut self, cmd: Cmd)` in this commit, I will try to implement this feature in future.

gwenn:
You need the `State` to invoke a `Cmd` on a `LineBuffer` for `Undo` / `Kill` ring.
Like `Changeset` here:
https://github.com/kkawakam/rustyline/blob/ceafb7f8057974201d0e76ac9fc7aaac3998a048/src/completion.rs#L102-L105

PR #819, opened 2024-10-21
Support for line continuation

aljazerzen:
Adds support for line continuation in prompts, which can be used to achieve rectangular prompts.

Supersedes #338 & #372

image

Review notes:
- I suggest reviewing commits separately,
- To check the thing out, run `input_multiline` example,
- The `breaking:` commit is the only breaking change that adds `continuation` param to `fn readline`. I'm not fully confident that this feature is worth breaking the API over, but it seems most reasonable public API to me. That's because the continuation is usually dependent on the prompt.
- Alternative public API would `fn Editor::set_continuation` function. In that case, I can revert last two commits.

I'm very invested in landing this PR and I'm willing to make changes.

TODO:

- [ ] windows support
- [ ] documentation
- [ ] tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ fn main() -> Result<()> {
         println!("No previous history.");
     }
     loop {
-        let readline = rl.readline(">> ");
+        let readline = rl.readline(">> ", "");
         match readline {
             Ok(line) => {
                 rl.add_history_entry(line.as_str());
--- a/examples/custom_key_bindings.rs
+++ b/examples/custom_key_bindings.rs
@@ -101,7 +101,7 @@ fn main() -> Result<()> {
     );
 
     loop {
-        let line = rl.readline("> ")?;
+        let line = rl.readline("> ", "")?;
         rl.add_history_entry(line.as_str())?;
         println!("Line: {line}");
     }
--- a/examples/diy_hints.rs
+++ b/examples/diy_hints.rs
@@ -90,7 +90,7 @@ fn main() -> Result<()> {
     rl.set_helper(Some(h));
 
     loop {
-        let input = rl.readline("> ")?;
+        let input = rl.readline("> ", "")?;
         println!("input: {input}");
     }
 }
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -73,7 +73,7 @@ fn main() -> rustyline::Result<()> {
     loop {
         let p = format!("{count}> ");
         rl.helper_mut().expect("No helper").colored_prompt = format!("\x1b[1;32m{p}\x1b[0m");
-        let readline = rl.readline(&p);
+        let readline = rl.readline(&p, "");
         match readline {
             Ok(line) => {
                 rl.add_history_entry(line.as_str())?;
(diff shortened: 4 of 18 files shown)


gwenn:
- continuation prompt should be optional (or behind a feature). Most likely a new `Prompt` trait with a default impl for &str.
- display size <> `continuation.len()`
- see #795 for impact of continuation prompt on highlighting
- see https://github.com/gwenn/rustyline/pull/2 for impact on layout
- see https://github.com/gwenn/rustyline/pull/3#issuecomment-2382345060 for another impact (and an alternative / similar PR made by @zao111222333)
- and another advice here: https://github.com/kkawakam/rustyline/pull/793#issuecomment-2286670046

This feature is a minefield...

aljazerzen:
> Most likely a new Prompt trait with a default impl for &str.

I like this. It would also avoid the breaking change.

---

Thank you for the linked info. Given that I manage to solve/avoid these problems, is this a wanted feature?

I don't want to spend a lot of time working on a feature that would not be merged due to big-picture reasons.

gwenn:
> Thank you for the linked info. Given that I manage to solve/avoid these problems, is this a wanted feature?
> 
> I don't want to spend a lot of time working on a feature that would not be merged due to big-picture reasons.

This is definitely a wanted feature.
But that raises many issues on the current implementation (Highlighting / Layout / Moving cursor without repainting everything / ...).
So you should probably use `reedline` which already supports continuation prompt instead of `rustyline`.

aljazerzen:
Ok.

Another question: is current behavior of `move_to_line_up` & friends desired behavior?
I observe the following behavior:
If I have:
- `prompt = "> "`
- `line = "hello\wor"`
- cursor at the end of line
- ```
  > hello
  wor_
  ```
- when I press arrow up, cursor moves after first `l`,
- ```
  > hel_o
  wor
  ```

I would expect it to move after `h`:

```
> h_llo
wor
```

I believe this needs to be fixed, right? And you've attempted to fix it in https://github.com/gwenn/rustyline/pull/2 ?

gwenn:
> I believe this needs to be fixed, right? And you've attempted to fix it in [gwenn#2](https://github.com/gwenn/rustyline/pull/2) ?

Current behavior might be buggy.
And https://github.com/gwenn/rustyline/pull/2 is just a draft to make this method work with continuation prompt.
We want something that can tell to the `LineBuffer`, the (line, column) associated to one byte / grapheme offset and vice-versa.

zao111222333:
You can have a look at https://github.com/gwenn/rustyline/pull/3#issue-2529487951 , however it's mightly that the Layout requires reconstruction, and the continuation feature seems need to implement base on the new Layout.

aljazerzen:
I've opened a new PR that implements span tracking first: #820

aljazerzen:
Closing this as it seems like maintainers are not keen on the implementation approach and want to expand the scope of the feature / future-proof the impl.

aljazerzen:
For anyone finding this, I'd recommend reedline instead.

PR #834, opened 2024-12-14
Use is_unsupported_term on both unix and windows

### LINKED ISSUE #832: Send prompt to stdout even if stdin is not a tty or check TERM also on Windows
This issue arises from the investigation in https://github.com/mthom/scryer-prolog/issues/2694. There is a comment at the end explaining my conclusions but I will re-word it here below.

I was having issues calling Scryer Prolog from emacs. Scryer Prolog uses rustyline for the repl, and it will print a prompt on the repl (`?- `). The emacs package `ediprolog` expects that prompt for running.

On Linux, the emacs shell `eshell` is identified by rustyline as an unsupported terminal, as it sets `TERM=dumb`. There `rustyline` [flushes the prompt to stdout](https://github.com/kkawakam/rustyline/blob/5b92355961718474fb67a84ae1520cbd75368f3d/src/lib.rs#L653-L660).

On Windows, the emacs shell `eshell` also sets `TERM=dumb` but this is not checked (it is said that [checking `TERM` on Windows is a no-op](https://github.com/kkawakam/rustyline/blob/5b92355961718474fb67a84ae1520cbd75368f3d/src/tty/windows.rs#L722-L725), not sure why) so it falls into [the third case of `readline_with`](https://github.com/kkawakam/rustyline/blob/5b92355961718474fb67a84ae1520cbd75368f3d/src/lib.rs#L673-L677) which does not flush the prompt.

I think at least with MSYS2 and others it is becoming more common for Windows applications to use things like `TERM=dumb` so I guess the check for "unsupported terminals" could be run also on Windows?

gwenn:
Fix #832

### DIFF
--- a/src/tty/mod.rs
+++ b/src/tty/mod.rs
@@ -1,5 +1,8 @@
 //! This module implements and describes common TTY methods & traits
 
+/// Unsupported Terminals that don't support RAW mode
+const UNSUPPORTED_TERM: [&str; 3] = ["dumb", "cons25", "emacs"];
+
 use crate::config::{Behavior, BellStyle, ColorMode, Config};
 use crate::highlight::Highlighter;
 use crate::keys::KeyEvent;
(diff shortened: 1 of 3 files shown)


gwenn:
At least no regression with `Powershell`.

jasagredo:
Unfortunately Scryer prolog is using rustyline 14.0.0 and your changes do not cherry-pick cleanly there. I can confirm that it works with the following patch I crafted when I opened the issue:
[windows-term.patch](https://github.com/user-attachments/files/18157501/windows-term.patch)

The changes are fairly similar to those in your branch, but yours look more organized (mine have duplicated `UNSUPPORTED_TERMS` and so on) so I'm pretty sure it will work, albeit I have not really checked it.

Issue #826, opened 2024-11-16
Interpreting graphene clusters when calculating width breaks on most terminal emulators


qsantos:
At least on Unix, when calculating the width of displayed characters, [rustline uses graphene segmentation](https://github.com/kkawakam/rustyline/blob/bc06b3eb322df1a535e6e3b64db8c0fee0b1df8c/src/tty/unix.rs#L1054).

However, using the [minimal example](https://github.com/kkawakam/rustyline/blob/bc06b3eb322df1a535e6e3b64db8c0fee0b1df8c/examples/minimal.rs) and pasting `👨‍👩‍👧‍👦` (`\u{1f468}\u{200d}\u{1f469}\u{200d}\u{1f467}\u{200d}\u{1f466}`), and then typing `A`, results in the following output:
![image](https://github.com/user-attachments/assets/60403c3c-71d1-437d-87cc-ea3edca28554)

This is because my terminal does not interpret the ZERO WIDTH JOINER (U+200D). In fact, I was able to reproduce this behavior in the following terminal emulators:

- xfce4-terminal
- gnome-terminal
- rxvt-unicode
- mate-terminal
- blackbox-terminal
- Putty
- Kitty
- Mac's Terminal
- iTerm2
- VS Code's built-in terminal
- [Intellij IDEA's terminal](https://stackoverflow.com/questions/64598298/zwj-zero-width-joiner-sequence-emoji-showing-as-2-emoji)

Unicode does say that the full graphene should be considered, and [unicode-width](https://github.com/unicode-rs/unicode-width) implement it so:

```rust
use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};

fn main() {
    let s = "\u{1f468}\u{200d}\u{1f469}\u{200d}\u{1f467}\u{200d}\u{1f466}";
    println!("{} {s}", s.width());
    for c in s.chars() {
        println!("{} {c}", c.width().unwrap_or(9));
    }
}
```

outputs:

```
2 👨‍👩‍👧‍👦
2 👨
0 ‍
2 👩
0 ‍
2 👧
0 ‍
2 👦
```

The first line looks correct in a graphical browser, but this is what I actually see:

![image](https://github.com/user-attachments/assets/080b1ecf-10f8-402e-b492-708972eb24fb)

Also note that this is not about legacy vs extended graphemes. ZERO WIDTH JOINER is considered in both.

[...]

PR #838, opened 2025-01-15
Grapheme clusters

gwenn:
See #826

- [X]  At least, make the mode configurable
- [X] Ideally, try to auto-detect / auto-select the mode based on some heuristics (but it seems difficult)

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -13,6 +13,23 @@ pub(crate) fn swidth(s: &str) -> Unit {
     Unit::try_from(s.width()).unwrap()
 }
 
+fn wcwidth(s: &str) -> Unit {
+    let mut width = 0;
+    for c in s.chars() {
+        width += cwidh(c);
+    }
+    width
+}
+
+const ZWJ: char = '\u{200D}';
+fn no_zwj(s: &str) -> Unit {
+    let mut width = 0;
+    for x in s.split(ZWJ) {
+        width += swidth(x);
+    }
+    width
+}
+
 #[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
 pub struct Position {
     pub col: Unit, // The leftmost column is number 0.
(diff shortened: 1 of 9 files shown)


src/layout.rs
```diff
@@ -1,18 +1,88 @@
...
+
+    /// Use environment variables to guess current mode
+    #[cfg(not(test))]
+    pub fn from_env() -> Self {
+        let gcm = match std::env::var("TERM_PROGRAM").as_deref() {
+            Ok("Apple_Terminal") => GraphemeClusterMode::Unicode,
+            Ok("iTerm.app") => GraphemeClusterMode::Unicode,
+            Ok("WezTerm") => GraphemeClusterMode::Unicode,
+            Err(std::env::VarError::NotPresent) => match std::env::var("TERM").as_deref() {
+                Ok("xterm-kitty") => GraphemeClusterMode::NoZwj,
```
qsantos, line 29: It is almost perfect. The one issue I had was with Kitty (on Debian):

Image

This seems to work better with:

```suggestion
                Ok("xterm-kitty") => GraphemeClusterMode::Unicode,
```

gwenn: Maybe the rendering varies depending on the platform / font used by Kitty ?
Because, I am almost sure that I tested kitty on MacOS. I will double check.
And, on Windows, all seems wrong:
```
Alacritty ~ (TERM=xterm-256color)
contour ~ (TERMINAL_NAME=contour)
WezTerm ~
Windows Terminal ~
```
But anyway, thanks for your feedback.

gwenn:
Wezterm is almost good on Windows:
image

src/layout.rs
gwenn: On MacOS, kitty seems fine:
image
(at least rustyline behaves like zsh)

qsantos: I tried changing the font for Kitty, but my understanding is that it just automatically use a fallback font for emojis if it is not present in the provided font.

I suspect that the issue I have observed does depend on the installed font, and I do not see how this could be handled 100% of the time. When testing in `xfce4-terminal`, My guess is that the underlying terminal emulator (GNOME’s [`vte`](https://github.com/GNOME/vte)) does something that prevent the combination of the separate emojis, regardless of the font. It is unclear whether that is intentional, but it does mean that the behavior stays consistent accross installations.

Anyway, having a bunch of empty space next to the character is not a terrible failure case. It is probably better than the original behavior where it was hard to know what you were even typing. So, probably a worthwile trade-off.

PR #839, opened 2025-01-17
Added noline to comparison

eivindbergem:
`noline` is a `no_std` line editor written in Rust. I'm the maintainer, so this is shameless promotion on my part.


gwenn:
Thanks.

PR #854, opened 2025-03-31
Add an option to show list completion options after a single tab

### LINKED ISSUE #853: Option for single-tab completion when using CompletionType::List?
The description of`CompletionType::List` doesn't seem to match its behaviour:

```rust
    /// Complete till longest match.
    /// When more than one match, list all matches
    /// (like in Bash/Readline).
```

What happens in practice is that when there's more than one match, the longest common prefix (LCP) is completed, and then the list is only shown if tab is pressed a second time.

This is rather confusing if you have the bell disabled, because the first tab press does nothing if the LCP has already been reached, with no indication that a second tab is required.

Would a PR that adds an option to show the list immediately be accepted?

irh:
Resolves #853.

Based on a patch proposed by @gwenn.


examples/example.rs
```diff
@@ -52,7 +52,8 @@ fn main() -> rustyline::Result<()> {
     env_logger::init();
     let config = Config::builder()
         .history_ignore_space(true)
-        .completion_type(CompletionType::List)
+        .completion_type(CompletionType::Fuzzy)
```
gwenn, line 55: Please, don't use Fuzzy.

irh: Oh whoops sorry, this was an experiment that wasn't supposed to go in the PR.

gwenn:
Thanks.

PR #856, opened 2025-04-09
Update README.md example code

yeoleobun:
Missing following `?` or leading `let _ = ` , will get warning "unused `Result` that must be used"


gwenn:
Thanks.

PR #842, opened 2025-01-26
Add support to SIGINT

gwenn:
See #780, #804, #835

- [X] finalize draft,
- [X] tests
- [ ] `signal-hook` impl,
- [ ] ...

PR #864, opened 2025-05-13
Fix SIGWINCH for signal-hook

krfricke:
With `features = ["signal-hook"]`, rustyline currently panics when the window is resized.

Repro:

```
use rustyline::DefaultEditor;
use rustyline::Result;

fn main() -> Result<()> {
    let mut rl = DefaultEditor::new()?;

    loop {
        let line = rl.readline("")?;
        rl.add_history_entry(line.as_str())?;

        println!("Line: {line}");
    }
}
```

Compile with `signal-hook` feature.

Run and resize window (on linux/unix).

This is because signal_hook sends `X` to the pipe, but we only process `I`/`W`, i.e. the custom signals we produce in the non-signal-hook feature.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -89,6 +89,8 @@ impl Signal {
         match b {
             b'I' => Self::Interrupt,
             b'W' => Self::Resize,
+            #[cfg(feature = "signal-hook")]
+            b'X' => Self::Resize,
             _ => unreachable!(),
         }
     }
--- a/src/error.rs
+++ b/src/error.rs
@@ -85,14 +85,18 @@ pub enum Signal {
 
 #[cfg(unix)]
 impl Signal {
+    #[cfg(not(feature = "signal-hook"))]
     pub(crate) fn from(b: u8) -> Self {
         match b {
             b'I' => Self::Interrupt,
             b'W' => Self::Resize,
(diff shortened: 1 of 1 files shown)


gwenn:
Bug introduced by https://github.com/kkawakam/rustyline/pull/842
But we cannot use `b'X'`:
https://docs.rs/signal-hook/latest/signal_hook/low_level/pipe/index.html
> Whenever there’s a signal, the signal handler writes **one byte of garbage** data to the write end

So a fix should be something like:
```patch
diff --git a/src/error.rs b/src/error.rs
index 87a844a..caec0c8 100644
--- a/src/error.rs
+++ b/src/error.rs
@@ -85,6 +85,12 @@ pub enum Signal {

 #[cfg(unix)]
 impl Signal {
+    #[cfg(feature = "signal-hook")]
+    pub(crate) fn from(_: u8) -> Self {
+        Self::Resize
+    }
+
+    #[cfg(not(feature = "signal-hook"))]
     pub(crate) fn from(b: u8) -> Self {
         match b {
             b'I' => Self::Interrupt,
```

krfricke:
Fair point, updated the PR

gwenn:
Thanks.

Issue #427, opened 2020-08-07
 Common information sharing for helper components


suikabreaker:
It would be helpful for different components of helper, e.g. , completer and hinter, to share some information.

Completer and hinter may both take item from a candidate list.

Furthermore, they may share the intermidiate parsing result to avoid parsing input multiple times and replicating calls to parsing code.

gwenn:
See:
https://github.com/kkawakam/rustyline/blob/bc687690309a0c4cd4f67fbff8f7569250b816c2/BUGS.md#document--syntax
> Because, when you have tokens (which may be as simple as words) or an AST, completion / suggestion / highlighting / validation become easy. So we need to send events to a lexer/parser, update Document accordingly. And fix Completer / Hinter / Highlighter API such as they have access to Document.
https://github.com/kkawakam/rustyline/issues/207#issuecomment-503204915
> For one syntax, Hinter, Highlighter and Completer, Validator should/would/will be correlated.
Because they should/would/will share the same Tokenizer/Parser.
Currently, each one parses the same input...

PR #866, opened 2025-05-18
Introduce Parser trait

gwenn:
- <https://github.com/kkawakam/rustyline/blob/master/BUGS.md#document--syntax>
- <https://github.com/kkawakam/rustyline/issues/427>
- <https://github.com/gwenn/rustyline-notes/blob/master/src/syntax.md>
- ...


gwenn:
Remove `Highlighter` from `Renderer::refresh_line`: use it directly from caller and pass styled `&str` to `Renderer::refresh_line`

gwenn:
`rustyline-derive` needs to be fixed.
And then we need a concrete example / test to validate the API. Maybe `sqlite3-parser` ?

PR #888, opened 2025-08-22
Add "replace and move" command

### LINKED ISSUE #883: New command for "replace and move"
It is currently impossible to create a keybinding to replace text in the line, and then move the cursor position (for example, to the end of the replacement text).

In the absence of a mechanism to return multiple commands from an event handler (as described in #306) I propose a smaller change, to create a new `Cmd` enum value `ReplaceAndMove`, based on the existing `Replace`:

```
/// vi-change-to, vi-substitute
Replace(Movement, Option<String>),
/// vi-change-to, vi-substitute
ReplaceAndMove(Movement, Option<String>, Movement),
```

This will also require refactoring the `execute<H: Helper>()` function, which currently has a big `match` block containing separate cases for `Cmd::Move(...)`.  I propose to move those cases to a new `execute_move<H: Helper>()` function, taking the `Movement` value:

```
match cmd {
    ...
    Cmd::Move(mvt) => {
        execute_move(s, mvt)?;
    }
    ...
    Cmd::ReplaceAndMove(mvt1, text, mvt2) => {
        s.edit_kill(&mvt1, kill_ring)?;
        if let Some(text) = text {
            s.edit_insert_text(&text)?;
        }
        execute_move(s, mvt2)?;
    }
    ...
}
```

hdwalters:
Fixes #883.

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -44,19 +44,8 @@ pub fn execute<H: Helper>(
         Cmd::Insert(n, text) => {
             s.edit_yank(input_state, &text, Anchor::Before, n)?;
         }
-        Cmd::Move(Movement::BeginningOfLine) => {
-            // Move to the beginning of line.
-            s.edit_move_home()?;
-        }
-        Cmd::Move(Movement::ViFirstPrint) => {
-            s.edit_move_home()?;
-            if s.line.starts_with(char::is_whitespace) {
-                s.edit_move_to_next_word(At::Start, Word::Big, 1)?;
-            }
-        }
-        Cmd::Move(Movement::BackwardChar(n)) => {
-            // Move back a character.
-            s.edit_move_backward(n)?;
+        Cmd::Move(mvt) => {
+            execute_move(s, mvt)?;
         }
         Cmd::ReplaceChar(n, c) => s.edit_replace_char(c, n)?,
         Cmd::Replace(mvt, text) => {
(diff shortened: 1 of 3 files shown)


hdwalters:
Hi @gwenn, I would like to get this pull request reviewed for merging into the main branch, but I'm not sure how to go about this. I would request a specific reviewer via GitHub, but cannot because I am not on the list of official contributors for the project.  Can you please advise?

gwenn:
Sorry but you have chosen the easy way by adding another command (instead of trying to support multiple commands).
And you have ignored undo:
https://github.com/kkawakam/rustyline/blob/master/CustomBinding.md#multiple--complex-actions
> We should ask the undo manager to start a "transaction" before first action and commit it after the last action. 

And I am biased but your use case looks fishy to me:
https://stackoverflow.com/questions/79731513/how-to-position-cursor-after-replaced-text-with-conditional-event-handler-in-ru
>  to add a space after every unique word completion, so I can continue entering SQL keywords afterwards.
> In a different context, I want to enter a column name after the table name. However, if there is already a space after the table,

Spaces should be meaningless in SQL:
```sh
sqlite> SELECT sqlite_schema . name FROM sqlite_schema;
-- no error 
```
And maybe you should not add space automatically after completion (I don't know any SQL line-editor that behaves like this)...

Issue #418, opened 2020-07-23
Execute Arbitrary Command Via Keybinding


fdncred:
Since we have introduced keybindings in `nushell` we're getting some interest in using `fzf` since `skim` doesn't work on Windows. Along with that comes the ability to execute any arbitrary program via those configured keybindings. Is it possible to 'shell out' and execute a command that is defined in a keybinding today? I can't find it if it is. If it isn't can you point me to the location where this could be inserted via  PR?

To be clear, this is really a two-pronged question.
1. we'd like to be able to use fzf somehow in mac/*nix/windows
2. we'd like to be able to run an executable via a keybinding.

Thanks.

gwenn:
Do you know a library (in any language) that already supports theses features ?

> 1. we'd like to be able to use fzf somehow in mac/*nix/windows

Could you please confirm that you want something like this:
https://github.com/kkawakam/rustyline/issues/302
> Maybe we should enhance the Completer trait and let implementations do their own complete_line logic.

https://github.com/kkawakam/rustyline/blob/master/src/completion.rs#L52
```rust
pub trait Completer {
...
    fn complete_line(&self
    someContext: SomeContext, // to be defined
    config: &Config,
) -> Result<Option<Cmd>> {
      // current default behaviour (redefined by nushell)
    }
}
```
So rustyline let nushell customize the display of candidates and user interaction.
But during the call to `complete_line`:
* terminal should be kept in raw mode (or at least rollbacked to raw mode before complete_line returns),
* rustyline will not be able to keep the screen coherent (rustyline repaints only its own lines).

2. we'd like to be able to run an executable via a keybinding.

Could you please confirm that you want something like this:
https://github.com/kkawakam/rustyline/pull/342
which may not interact well with the undo manager...

jonathandturner:
@gwenn - I'm coming up to speed on the current threads, but these are my early thoughts:

for #302 - I think what we want is more than just tabbing, but it's a start. Part of the problem here is that rustyline supports skim, but skim doesn't support Windows. We've been holding off adding support in nushell until the Windows support is available. Being able to do our own completions does have some advantages, so we could optionally call out into an external command like `fzf`.

Right, there'd need to be some contract between rustyline and the code handling the completion that things come back in a state that rustyline can handle.

[...]

PR #895, opened 2025-09-08
Add Cmd::Macro for keystroke replay functionality

ticpu:
This commit implements a new Cmd::Macro variant that allows replaying sequences of keystrokes character-by-character, enabling function key bindings to execute commands automatically.

Since this is replaying a command, it integrates with the command processing pipeline seamlessly, without allowing for completely arbitrary commands to be execute.

Example usage:
```rust
// Bind F1 to execute "help" command automatically
rl.bind_sequence(
    KeyEvent(KeyCode::F(1), Modifiers::NONE),
    Cmd::Macro("help\n".to_string())
);
```

When triggered, the macro replays each character as if the user typed it, and the trailing newline triggers AcceptLine to submit the command automatically.

It is related to issue #418: Execute Arbitrary Command Via Keybinding, I took insipration from that issue's name as I needed to "execute" arbitrary command (and not executable) in my application.

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -47,6 +47,18 @@ pub fn execute<H: Helper, P: Prompt + ?Sized>(
         Cmd::Macro(macro_str) => {
             s.start_macro(macro_str);
         }
+        Cmd::MacroClearLine(macro_str) => {
+            // Save the current line content for restoration after macro execution
+            if !s.line.is_empty() {
+                let line_content = s.line.as_str().to_string();
+                s.macro_player_mut().set_pending_restore(line_content);
+            }
+            // Clear the current line
+            s.edit_kill(&Movement::WholeLine, kill_ring)?;
+            s.changes.end();
+            // Then start the macro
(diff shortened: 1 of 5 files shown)


gwenn:
See https://github.com/kkawakam/rustyline/blob/master/CustomBinding.md#multiple--complex-actions

ticpu:
@gwenn Hah, that's interesting, thanks for sharing.

I think both could co-exist in this case. Macro is addressing a couple use-cases by simulating the user's input; so it can't really get "deprecated"; I see the potential of adding the more complete Custom Bindings but right now, the only alternative would be to move to a different library until it is fully implemented.

I had to write this so I could have a feature-complete fs_cli (FreeSWITCH's CLI).

Would you see any issues to be able to do it both ways?  It could even use the custom bindings architecture by converting the text to actions later on.

ticpu:
Is there any chance of this being merged as a separate feature?

PR #949, opened 2026-06-23
Fix page_completions

### LINKED ISSUE #948: cursor misplaced after displaying a long completion suggestion list
when rustyline displays a long list of completions *after* asking the user whether they want to see all completions the cursor is placed in the wrong row. this can be reproduced on the example by completing a large directory, e.g. `/dev`:

Image

while ultimately benign this is quite confusing, and the last row of suggestions will be overwritten completely on the next input.

we suspect that the layout manipulation here: https://github.com/kkawakam/rustyline/blob/8cb7a468e42784edb4c366521454fc85a35bb6e5/src/lib.rs#L184 is incorrect. dropping it does fix the issue. not sure that's the correct fix though, so we're opening this as an issue instead of as a PR.

tested on rustyline 18.0 and current master

gwenn:
Fix #948

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -178,10 +178,10 @@ fn complete_line<H: Helper, P: Prompt + ?Sized>(
         s.edit_move_end()?;
         s.line.set_pos(save_pos);
         // we got a second tab, maybe show list of possible completions
-        let show_completions = if candidates.len() > config.completion_prompt_limit() {
+        let asked = candidates.len() > config.completion_prompt_limit();
+        let show_completions = if asked {
             let msg = format!("\nDisplay all {} possibilities? (y or n)", candidates.len());
             s.out.write_and_flush(msg.as_str())?;
-            s.layout.end.row += 1;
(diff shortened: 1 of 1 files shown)


pennae:
works perfectly now, thanks for the quick response! :)

is there any chance of getting this fix as a 18.1 release? (no worries if not, we can keep a fork for now. unfortunately master won't build on our rust targets any more due to assert_matches things, so we may end up needing a fork anyway until we can upgrade those toolchains)

gwenn:
> is there any chance of getting this fix as a 18.1 release?

I am afraid that I need to release version 19.0 because some changes already merged in master increase the MSRV.
(If necessary, I can still do a 18.1 release with only this PR but it takes more time)

pennae:
we're in no rush, it'll be a bit still before we can ship anything that uses rustyline at all. we would greatly appreciate a release, but we're also totally fine with keeping a local patch for a few months months (until nixpkgs stable gets a rustc 1.96 or newer). do as you think is best, we can work just as easily with an 18.1 release or a local patch :)

PR #964, opened 2026-08-19
Accept Tab as confirmation at the completion list prompt

### LINKED ISSUE #955: allow using `TAB` for "Display all X possibilities? (y or n)" prompt
when prompting the user for `Display all X possibilities?`, it'd be nice if <kbd>TAB</kbd> was treated as a "yes". this allows expanding the completion list without having to move your hand at all, slightly reducing friction.

for example, this is what zsh's line editor does (and i've gotten so used to it i forgot it isn't universal ^^;)

ChrisJr404:
Fixes #955. At the "Display all N possibilities? (y or n)" prompt, this accepts Tab as a "yes", so you can expand the completion list by pressing Tab again instead of reaching for the y key.

The prompt still waits for a keypress rather than consuming the Tab that opened it, and every other key keeps its current behavior. I added a test that drives `complete_line` with two Tabs and checks the list is shown.


src/test/mod.rs
```diff
@@ -100,6 +100,51 @@ fn complete_symbol() {
...
+        _pos: usize,
+        _ctx: &Context<'_>,
+    ) -> Result<(usize, Vec<String>)> {
+        Ok((0, vec!["rust".to_owned(), "rush".to_owned()]))
+    }
+}
+impl Hinter for AmbiguousCompleter {
+    type Hint = String;
+
+    fn hint(&self, _line: &str, _pos: usize, _ctx: &Context<'_>) -> Option<Self::Hint> {
```
gwenn, line 119: Cosmetic: you don't need to redefine this method.

PR #970, opened 2026-08-24
Fix EOF key binding on platforms where VMIN aliases VEOF

plotnick:
`enable_raw_mode` built its key map from the raw termios after setting `VMIN` and `VTIME`. But on some platforms (illumos & other Solaris derivatives, linux-sparc64, AIX, and Haiku, according to nix aliasing tables), `VMIN` and `VTIME` share the `c_cc` slots of `VEOF` and `VEOL`, so the map bound the just-written `VMIN` value (`0x01`, Ctrl-A) to `EndOfFile` and the real EOF character did nothing. Read the special characters from the original mode instead, in both backends. Verified on illumos: Ctrl-D now ends the session, Ctrl-A moves to the cursor to the start of line as intended.

Offered as two commits: the first is the minimal correctness fix, the second is a refactored and testable version that I recommend, but leave to your discretion.

:robot: Disclaimer: found & fixed with LLM assistance, reviewed & validated with human :eyes: & :brain:.


gwenn:
Could you please try the `termios` feature instead ?
See #731

plotnick:
> Could you please try the `termios` feature instead ? See #731

We're already using the `termios` feature, that's the configuration where we found this (and the verification in the PR description was run with it). It isn't the constants problem from #731; the constants are correct in both crates. This is an ordering bug in `enable_raw_mode` itself: the key map is read out of the struct after `VMIN`/`VTIME` are written into it, but on these platforms those writes land in the `VEOF`/`VEOL` slots because they're shared. From `termios`'s `src/os/illumos.rs`:

```rust
pub const VEOF:     usize = 4;
pub const VEOL:     usize = 5;
...
pub const VMIN:     usize = 4;
pub const VTIME:    usize = 5;
```

So the `termios` backend clobbers `VEOF` identically, which is why the fix is the same for both backends (read the special characters from the original mode instead of the modified copy).

gwenn:
> We're already using the `termios` feature

https://github.com/kkawakam/rustyline/pull/970/changes#diff-bc8fd5ba686c57cce3f8eefc70d2b5419468e9c84b5d1923095584f8dabb010bL1603-L1604
```rust
#[cfg(not(feature = "termios"))]
mod termios_ {
```

Your patch says the opposite...

gwenn:
Both [termwiz](https://github.com/search?q=repo%3Awezterm%2Fwezterm+VEOF&type=code)
And [fish-shell](https://github.com/search?q=repo%3Afish-shell%2Ffish-shell+VEOF&type=code)
are using the `libc` directly.
Maybe we should do the same.

plotnick:
> ```rust
> #[cfg(not(feature = "termios"))]
> mod termios_ {
> ```
> 
> Your patch says the opposite...

It includes both cases, handled the same way. I can refactor the helper if you'd like, but it seemed like this preserved the symmetry that was already present.

gwenn:
Ok,
https://linux.die.net/man/3/termios
> The above symbolic subscript values are all different, except that VTIME, VMIN may have the same value as VEOL, VEOF, respectively. **In noncanonical mode the special character meaning is replaced by the timeout meaning.**

plotnick:
Yes, that's the aliasing. So this fix is independent of the termios layer: the special characters must be read before VMIN/VTIME are written, since the kernel only has the one array.

plotnick:
Verified your fix (5ab22f2) with our application on illumos, works great. Many thanks! I was tying my fingers in knots trying to remember to remap Ctrl-A to Ctrl-D on one platform but not the other.

PR #979, opened 2026-09-19
Allow custom bindings to override Emacs numeric prefixes

jph00:
In Emacs mode, Alt-digits and Alt-minus enter numeric-argument handling before custom bindings are checked. Applications therefore cannot bind these keys—for example, to insert mathematical symbols.

This checks custom bindings first. If no handler accepts the key, numeric-argument handling proceeds as before.

Adds a regression test for overriding Alt-0, Alt-3, Alt-9 and Alt-minus. The library suite passes, and I also tested the change through a real terminal in an application using conditional bindings.

(Prepared with AI help, reviewed by me personally.)


jph00:
@gwenn let me know if there's anything else you'd like me to test, or anything you'd prefer changed.

I'm using rustyline for the REPL of a new APL implementation and alt-digits is a critical piece of functionality expected for APL input.

